PR #1, opened 2016-12-29
Update garage.py

cyberking-saga:
Check for the case where initial and final states are given below:
initial = [3,1,2,0]
final = [2,3,0,1]

between state:
[3,0,2,1]
[0,3,2,1]
[2,3,0,1]
So only 3  moves needed while yourt program gives 5 moves.

### DIFF
--- a/array/garage.py
+++ b/array/garage.py
@@ -17,32 +17,33 @@
 # Each step swap with 0 only.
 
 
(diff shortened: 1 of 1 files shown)


keonkim:
brilliant!

cyberking-saga:
Thanks for merging.

PR #7, opened 2017-04-21
Update min_height.py

garykrige:


keon:
oops, i really gotta write the tests.
thanks @garykrige !

PR #14, opened 2017-04-23
added pythagoras.py to maths

stongaf:
added pythagoras.py to maths

### DIFF
--- a/math/pythagoras.py
+++ b/math/pythagoras.py
@@ -1,0 +1,16 @@
+"""
+input two of the three side in right angled triangle and return the third. use "?" to indicate the unknown side. 
+"""
+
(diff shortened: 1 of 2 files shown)


keon:
can you convert it to 4 spaces?
Thanks

math/pythagoras.py
```diff
@@ -0,0 +1,16 @@
+"""
+input two of the three side in right angled triangle and return the third. use "?" to indicate the unknown side. 
+"""
+
+def pythagoras(opposite,adjacent,hypotenuse):
```
ankit167: should also handle the case when user enters more than one side as '?'. The result cannot be tabulated in this case.

stongaf: This is now done.

PR #19, opened 2017-04-25
added a comment to let beginners know about the assert function in python

tejas-kr:


delirious-lettuce:
@tejas-kr ,

Is this really necessary?

tejas-kr:
yes the beginners don't know about this assert statement

mingrammer:
Well, I think it is unnecessary, because it is not about logic but just python grammar.

keon:
I believe it is safe to assume that users know what assert is :)

tejas-kr:
ok as you wish

PR #25, opened 2017-04-25
make_sentence.py renamed to make_sentece.cpp

dharan1011:
The code was in C++ format, but the file was saved as python file (i.e .py) so renamed as C++ extension (i.e .cpp).


keon:
the goal here was to convert it to python.
it was kind of TODO for me when this was just a private project.

Issue #24, opened 2017-04-25
How about adding links between README and Python code?


mingrammer:
It would be better to link the python source code and algorithm title on README.md. If you agree with this, I'm willing to add it and make PR.

keon:
absolutely bruh.
Sounds awesome.

delirious-lettuce:
@mingrammer ,

This sounds like a good idea.

PR #26, opened 2017-04-25
Make links between algorithm title and python source code

mingrammer:
I've made the all links between title and source code for #24 

1. Able to see the source code for all algorithms on README right away
2. Separate the `implementation` section and `design` section for readability
3. Hide some non-algorithm source files such as `__init__.py` and cache related files

You can squash the two commits with message as this PR title


keon:
Thanks!

PR #28, opened 2017-04-25
add algorithm to backtrack

vbyeh:
Added the solution to Word Search II on LeetCode under the backtrack directory

### DIFF
--- a/backtrack/word_search.py
+++ b/backtrack/word_search.py
@@ -1,3 +1,4 @@
+import unittest
 '''
 Given a matrix of words and a list of words to search, return a list of words that exists in the board
 This is Word Search II on LeetCode
@@ -11,11 +12,10 @@ board = [
 
 words = ["oath","pea","eat","rain"]
 
-print findWords(board, words)
 '''
 
-def findWords(board, words):
-    #make trie
+def find_words(board, words):
+    # make a trie structure that is essentially dictionaries of dictionaries that map each character to a potential next character
     trie = {}
     for word in words:
         curr_trie = trie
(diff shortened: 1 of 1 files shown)


backtrack/word_search.py
```diff
@@ -0,0 +1,49 @@
...
+         ['i','h','k','r'],
+         ['i','f','l','v']
+         ]
+
+words = ["oath","pea","eat","rain"]
+
+print findWords(board, words)
+'''
+
+def findWords(board, words):
```
keon: let's use snake case

backtrack/word_search.py
```diff
@@ -0,0 +1,49 @@
...
+    if i < 0 or i >= len(board) or j < 0 or j >= len(board[0]):
+        return
+    
+    if not used[i][j] and board[i][j] in trie:
+        used[i][j]=True
+        find(board,i+1,j,trie[board[i][j]],pre+board[i][j], used, result)
+        find(board,i,j+1,trie[board[i][j]],pre+board[i][j], used, result)
+        find(board,i-1,j,trie[board[i][j]],pre+board[i][j], used, result)
+        find(board,i,j-1,trie[board[i][j]],pre+board[i][j], used, result)
+        used[i][j]=False
```
keon: it'd be nice to have some test cases under if __name__ == "__main__".

backtrack/word_search.py
```diff
@@ -0,0 +1,49 @@
...
+
+    result = set()
+    used = [[False]*len(board[0]) for _ in range(len(board))]
+    
+    for i in range(len(board)):
+        for j in range(len(board[0])):
+            find(board, i, j, trie, '', used, result)
+    return list(result)
+
+def find(board, i, j, trie, pre, used, result):
```
keon: would 'backtract()' be a better name?

keon: also could you explain briefly what trie, pre, used and result are? (with types)

vbyeh:
thanks for the feedback! I changed a few things and added unittests

PR #30, opened 2017-04-26
added algorithm to reverse_string.py and minimized reverse_word.py

dharan1011:
I have added another pythonic techniques to reverse a string named pythonic2 and minimized the algorithm in reverse_word.py as it showed inconsistency :-1: 
```
// revese_word.py 
Old Output : apizz like I and kim keon am I
Updated Outpult : pizza like I and kim keon am I 
```

### DIFF
--- a/string/reverse_string.py
+++ b/string/reverse_string.py
@@ -19,7 +19,15 @@ def iterative(s):
 print(iterative(s))
 
 def pythonic(s):
-    r = list(reversed(s))
-    return "".join(r)
+	r = list(reversed(s))
+	return "".join(r)
 
 print(pythonic(s))
+
+def pythonic2(str):
+	res = ""
+	for s in str:
+		res = s + res
+	return res
+
+print(pythonic2(s))


Biswa369:
Hi

string/reverse_words.py
```diff
@@ -1,29 +1,9 @@
...
-        elif i == n-1:
-            if start:
-                reverse(arr, start, i)
-        else:
-            if not start:
-                start = i
-    return "".join(arr)
-
+def reverse_words(str):
+	res = " "
```
ankit167: We can further minimize this function-
res = ' '.join(str.split()[::-1])

dharan1011:
@keon isnt my request upto the mark?? :disappointed:

string/reverse_string.py
```diff
@@ -19,7 +19,15 @@ def iterative(s):
 print(iterative(s))
 
 def pythonic(s):
-    r = list(reversed(s))
-    return "".join(r)
+	r = list(reversed(s))
```
ankit167: Would recommend to skip this function, as it does not illustrate the algorithm behind reversing a string.

string/reverse_words.py
ankit167: The algorithm uses an extra variable to store the reversed string. A better algorithm would be to reverse the string in place.

ankit167:
@dharan1011 Please resolve conflicts and make a fresh request.

string/reverse_words.py
dharan1011: :+1: Yes ur suggestion seems right

PR #38, opened 2017-04-26
Added a Markov Chain implementation

probablytom:
Hope it's okay. I imagine it might not be simple enough, or it might be that the generator is too much -- curious to know whether it's okay. I think markov chains would be a good addition to this collection!


keon:
looks good!
Could we have some use cases (examples) and a brief explanation for the code?

probablytom:
Terribly sorry -- only just saw this! Are you looking for that here, or alongside the PR?

keon:
@probablytom sorry for a terribly late response.
I merged the PR. I can add the comments later :)

PR #39, opened 2017-04-26
Add extended GCD, Rabin-Miller primality test, and RSA encryption algorithm

nickolaswiebe:
`rsa.py` imports from `rabin_miller.py` and `extended_gcd.py`


nickolaswiebe:
All the problems should be fixed now.

PR #42, opened 2017-04-27
Fixed checks in 'if' conditions

ankit167:
Going by the current logic:
Input: I am keon kim and I like pizza
Output: apizz like I and kim keon am I

The first word 'apizz' is NOT correctly reversed to 'pizza'.
The reason being- when 'start' is '0', the 'if' block is not evaluated. 

After fix:
Input: I am keon kim and I like pizza
Output: pizza like I and kim keon am I


keon:
Thanks buddy,
we should consider a more interesting sentence next time :)

PR #43, opened 2017-04-27
Create counting_sort.py

quang2705:
Hey, I think this algorithm might be helpful

### DIFF
--- a/sort/counting_sort.py
+++ b/sort/counting_sort.py
@@ -1,0 +1,24 @@
+def couting_sort(arr,k):
+	"""Couting_sort
+		Sorting a array which has no element greater than k
+		Complexity: 0(n)
+	"""
+	temp_arr = [0]*(k+1)
+	for i in range(0,len(arr)):
+		temp_arr[arr[i]]= temp_arr[arr[i]]+1
+	#temp_array[i] is the times the number i appear in arr
+	print(temp_arr)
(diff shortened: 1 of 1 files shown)


sort/counting_sort.py
```diff
@@ -0,0 +1,24 @@
+def couting_sort(arr,k):
+	"""Couting_sort
+		Sorting a array which has no element greater than k
+		Complexity: 0(n)
+	"""
+	temp_arr = [0]*(k+1)
+	for i in range(0,len(arr)):
+		temp_arr[arr[i]]= temp_arr[arr[i]]+1
```
ankit167: Please use space before '='

sort/counting_sort.py
```diff
@@ -0,0 +1,24 @@
...
+	result_arr = [0]*len(arr)
+	#creating a result_arr an put the element in a correct positon
+	for i in range(len(arr)-1,-1,-1):
+		result_arr[temp_arr[arr[i]]-1] = arr[i]
+		temp_arr[arr[i]] = temp_arr[arr[i]]-1
+
+	return result_arr 
+
+
+x =couting_sort([2,5,3,0,2,3,0,3],9)
```
ankit167: shouldn't we pass 8 (length of the list) instead of 9? We are already doing k+1 inside the function

PR #51, opened 2017-05-03
Heap Sort

ihchen:
Noticed there wasn't an implementation of heap sort so decided to do one! Implemented two versions, one using a max heap and one using a min heap. Min heap is actually faster because it avoids a swap but I think the heapify code is a little less readable. Let me know which one should be final.

Also, this is my first PR! So let me know if I've done anything wrong in the fork/pull request/general GitHub process.

### DIFF
--- a/sort/heap_sort.py
+++ b/sort/heap_sort.py
@@ -2,7 +2,13 @@ def heap_sort(arr):
     """ Heapsort
         Complexity: O(n log(n))
     """
-    pass
+
+    for i in range(len(arr)-1,-1,-1):
+        heapify(arr, i)
+
+        temp = arr[0]
+        arr[0] = arr[i]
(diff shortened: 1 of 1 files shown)


sort/heap_sort.py
```diff
@@ -0,0 +1,89 @@
...
+            if arr[child+start] < arr[current_parent+start]:
+                temp = arr[current_parent+start]
+                arr[current_parent+start] = arr[child+start]
+                arr[child+start] = temp
+
+                current_parent = child
+            # If no swap occured, no need to keep iterating
+            else:
+                break
+
```
ankit167: Please wrap inputs and function calls inside __name__ == '__main__'

ihchen: Done. Sorry for the late response!

PR #55, opened 2017-05-06
Add another implementation to plusOne.py

Mohamed3on:
Solve the problem by converting the list to a string, then to an integer, then incrementing that integer and returning it as a list.

### DIFF
--- a/array/plus_one.py
+++ b/array/plus_one.py
@@ -36,3 +36,13 @@ def plus_one(digits):
     new_num = [0] * (n+1)
     new_num[0] = 1
     return new_num
+## another implementation, convert list to string, then to integer, then increment that integer
+def plusone(digits):
+        num=""
+        for d in digits:
+            num+=str(d)
+        incremented=str(int(num)+1)
+        incrementedAsList=[]
+        for i in incremented:
+            incrementedAsList.append(int(i))
+        return incrementedAsList    


array/plus_one.py
```diff
@@ -36,3 +36,13 @@ def plus_one(digits):
     new_num = [0] * (n+1)
     new_num[0] = 1
     return new_num
+## another implementation, convert list to string, then to integer, then increment that integer
+def plusone(digits):
+        num=""
+        for d in digits:
+            num+=str(d)
+        incremented=str(int(num)+1)
+        incrementedAsList=[]
```
venky18: you can convert  `string` to `list`  by  using `list`
so you can replace the last three lines as `incrementedAsList = list(incremented) `

Mohamed3on: Indeed, but it would convert it to a list of chars, while the needed return type was a list of ints, that's why I used a loop.

Mohamed3on: For reference:
https://leetcode.com/problems/plus-one/#/description

Mohamed3on:
Is there something wrong with the implementation? Or do you not want to add another implementation?

array/plus_one.py
keon: could you convert this to snake_case ?

ofek:
This solution is sub-optimal, but if accepted here is the condensed version:

```python
def plus_one(digits):
    return [int(s) for s in str(int(''.join(str(i) for i in digits)) + 1)]
```

ankit167:
@Mohamed3on Closing this request, as it has conflicts with master. Please push a fresh request.

PR #57, opened 2017-05-08
add find_missing_number

ofek:


ofek:
ready

bit/find_missing_number.py
```diff
@@ -0,0 +1,19 @@
+def find_missing_number(nums):
+    """Returns the missing number from a sorted list of integers
+    in O(n) time and space.
+
```
ankit167: What is the nature of the sorted list of integers here? Can difference between consecutive integers be more than 1?

ofek: No.

ankit167: I guess, we should mention this in the comments. 'Sorted list of integers' does not explain it.

ofek: Done.

ankit167:
what should be the output of input: [-1,0,2,3]? is it 1?

ofek:
@ankit167 Yes, thank you for checking edge cases. I've just added constraint of allowing only positive integers.

ankit167:
What if none of the numbers in a list are missing.
f.e. l = [0,1,2,3]. As per the code the output is 5, which I think is incorrect. Please handle this case, and other cases that you can think of.

ofek:
Fixed. A complete sequence now returns the next integer.

bit/find_missing_number.py
```diff
@@ -0,0 +1,15 @@
+def find_missing_number(nums):
+    """Returns the missing number from a sorted list of unique
```
ankit167: Since, we are doing xor operations here, I don't think it is required for the input list to be sorted, is it?
The code should work for a general list of integers in range [0..n], where one number is missing and the difference between any two numbers is 1.

ofek: Ah, yes ty.

PR #58, opened 2017-05-08
add bytes <-> int conversion

ofek:


bit/bytes_int_conversion.py
```diff
@@ -0,0 +1,32 @@
...
+
+
+def int_to_bytes_big_endian(num):
+    bytestr = deque()
+    while num > 0:
+        # list.insert(0, ...) is inefficient
+        bytestr.appendleft(num & 0xff)
+        num >>= 8
+    return bytes(bytestr)
+
```
ankit167: 2 blank spaces required between functions

ofek: Fixed.

PR #60, opened 2017-05-10
Added code to find the number of unique paths in a matrix

ankit167:
Starting from the top left cell, one can move right or bottom to reach bottom right cell of the matrix.
The number of unique paths are calculated.


ankit167:
@shruti19 Please review.

keon:
@ankit167 sorry! i am traveling... so I cannot respond quickly

PR #63, opened 2017-05-13
max product subarray problem: another approach 

shruti19:
Print max product as well as the subarray

### DIFF
--- a/dp/max_product_subarray_another_approach.rb
+++ b/dp/max_product_subarray_another_approach.rb
@@ -1,12 +1,11 @@
 def subarray_with_max_product(arr)
-	len = arr.length
+  	len = arr.length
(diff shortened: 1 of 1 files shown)


dp/max_product_subarray_another_approach.rb
```diff
@@ -0,0 +1,25 @@
+def subarray_with_max_product(arr)
```
ankit167: Please add a brief explanation before the function.

ankit167:
@shruti19 since, we are maintaining the code in Python, it would be better have the same approach written in Python (for uniformity)

ankit167:
Please append the new approach as a new function in dp/max_product_subarray.py, and send in a fresh request.

PR #75, opened 2017-05-24
Update buy_sell_stock.py

dong-jy:
I don't know exactly if I was correct.
Please review it.
Thanks.

### DIFF
--- a/dp/buy_sell_stock.py
+++ b/dp/buy_sell_stock.py
@@ -20,7 +20,21 @@ In this case, no transaction is done, i.e. max profit = 0.
 """
 
 
-def max_profit(prices):
+# O(n^2) time
+def max_profit_naive(prices):
+    """
+    :type prices: List[int]
+    :rtype: int
+    """
+    max_so_far = 0
+    for i in range(0, len(prices) - 1):
+        for j in range(i + 1, len(prices)):
+            max_so_far = max(max_so_far, prices[j] - prices[i])
+    return max_so_far
+
+
+# O(n) time
+def max_profit_optimized(prices):
     """
     input: [7, 1, 5, 3, 6, 4]
     diff : [X,-6, 4,-2, 3,-2]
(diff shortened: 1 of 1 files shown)


dp/buy_sell_stock.py
```diff
@@ -23,12 +23,13 @@
 def max_profit(prices):
     """
     input: [7, 1, 5, 3, 6, 4]
-    diff : [X,-6, 5,-2, 3,-2]
+    diff : [X,-6, 4,-2, 3,-2]
```
ankit167: Please use spaces around ','

dp/buy_sell_stock.py
```diff
@@ -23,12 +23,13 @@
...
     :type prices: List[int]
     :rtype: int
     """
     cur_max, max_so_far = 0, 0
     for i in range(1, len(prices)):
```
ankit167: should 'i' start from 0 and 'j' start from 'i+1'? Correct me if I am wrong.

ankit167:
@dong-jy Since, we are allowed to make only a single transaction, this problem is same as finding max difference between two numbers, such that the larger number appears after the smaller number. Your solution is correct. However, it takes O(n^2) time complexity. I believe, we can work this problem out in O(n) time. In case you are interested, please append existing code with the optimized version as well.

dp/buy_sell_stock.py
dong-jy: Thanks!
But what does 'nit' mean?
And I used space around ',', kept same format with original lines.

dp/buy_sell_stock.py
dong-jy: Got it.

dp/buy_sell_stock.py
ankit167: nit is the short form for 'nit pick'. It denotes a change which is not _important_, but is technically _correct_

I can see space missing before the negative numbers. (the reason I commented)

dong-jy: @ankit167 
 I just revised the code according to your review.
May you review it again if your are free? Or just edit or delete this PR is if doesn't meet the standard.

The original author just  wrote this way in comment (without space before minus numbers), so I follow him.

ankit167:
@dong-jy May be, we should also add algorithm to find the maximum profit, where multiple transactions (buy/sell) are allowed. That can be a separate pull request. Please take this one up as well. Thanks.

PR #86, opened 2017-06-12
Adding a cycle detection algorithm for graphs.

karpinski:
Adding an implementation and a respective entry in the README.

### DIFF
--- a/graph/cycle_detection.py
+++ b/graph/cycle_detection.py
@@ -44,7 +44,7 @@ def contains_cycle(graph):
     traversal_states = {vertex: TraversalState.WHITE for vertex in graph}
     for vertex, state in traversal_states.items():
         if (state == TraversalState.WHITE and
-            is_in_cycle(graph, traversal_states, vertex)):
+           is_in_cycle(graph, traversal_states, vertex)):
             return True
     return False
 
--- a/graph/cycle_detection.py
+++ b/graph/cycle_detection.py
@@ -1,0 +1,52 @@
+"""
+Given a directed graph, check whether it contains a cycle.
+
+Real-life scenario: deadlock detection in a system. Processes may be
+represented by vertices, then and an edge A -> B could mean that process A is
+waiting for B to release its lock on a resource.
+"""
+from enum import Enum
+
+
+class TraversalState(Enum):
+    WHITE = 0
+    GRAY = 1
+    BLACK = 2
+
+
+example_graph_with_cycle = {'A': ['B', 'C'],
+                            'B': ['D'],
+                            'C': ['F'],
+                            'D': ['E', 'F'],
+                            'E': ['B'],
+                            'F': []}
+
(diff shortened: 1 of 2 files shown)


Aluriak:
Nodes states should be indicated with an enum, not an undocumented integer value (or at least be documented).

This code uses side-effect on `traversal_state`. Should we implement a simple topological sort instead ?

graph/cycle_detection.py
```diff
@@ -0,0 +1,44 @@
...
+                               'B': ['D', 'E'],
+                               'C': ['F'],
+                               'D': ['E'],
+                               'E': [],
+                               'F': []}
+                               
+
+
+def is_in_cycle(graph, traversal_state, vertex):
+  if traversal_state[vertex] == 1:
```
ankit167: Plz keep indentation as a multiple of 4

karpinski: Done

karpinski:
@Aluriak I added an enum for determining node traversal states. As for your second suggestion, I think that since this repository is supposed to contain "minimal examples" of algorithms, implementation of a simple DFS for cycle detection is as minimal as it gets. At the same time, I think that it would be good to have a separate topological sort implementation in the repo.

graph/cycle_detection.py
```diff
@@ -0,0 +1,52 @@
...
+            return True
+    traversal_states[vertex] = TraversalState.BLACK
+    return False
+
+
+def contains_cycle(graph):
+    traversal_states = {vertex: TraversalState.WHITE for vertex in graph}
+    for vertex, state in traversal_states.items():
+        if (state == TraversalState.WHITE and
+           is_in_cycle(graph, traversal_states, vertex)):
```
keon: why 3 spaces ?

graph/cycle_detection.py
```diff
@@ -44,7 +44,7 @@ def contains_cycle(graph):
     traversal_states = {vertex: TraversalState.WHITE for vertex in graph}
     for vertex, state in traversal_states.items():
         if (state == TraversalState.WHITE and
-            is_in_cycle(graph, traversal_states, vertex)):
+           is_in_cycle(graph, traversal_states, vertex)):
```
keon: why 3 spaces?? @ankit167

ankit167: @keon Added this commit to avoid the following pep8 error, which i was getting while running the style checks:
**visually indented line with same indent as next logical line**

As per pep8,
When the conditional part of an if -statement is long enough to require that it be written across multiple lines, it's worth noting that the combination of a two character keyword (i.e. if ), plus a single space, plus an opening parenthesis creates a natural 4-space indent for the subsequent lines of the multiline conditional. This can produce a visual conflict with the indented suite of code nested inside the if -statement, which would also naturally be indented to 4 spaces.

PR #87, opened 2017-06-13
Adding an implementation for the knapsack problem.

karpinski:
Adding the implementation and a respective README entry.

### DIFF
--- a/dp/knapsack.py
+++ b/dp/knapsack.py
@@ -1,0 +1,38 @@
+"""
+Given the capacity of the knapsack and items specified by weights and values,
+return the maximum summarized value of the items that can be fit in the
+knapsack.
+
+Example:
+capacity = 5, items(value, weight) = [(60, 5), (50, 3), (70, 4), (30, 2)]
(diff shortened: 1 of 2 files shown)


dp/knapsack.py
```diff
@@ -0,0 +1,37 @@
...
+
+def get_maximum_value(items, capacity):
+    dp = [0] * (capacity + 1)
+    for item in items:
+        for current_weight in range(capacity + 1):
+            total_weight = current_weight + item.weight
+            if total_weight <= capacity:
+                dp[total_weight] = max(dp[total_weight],
+                                       dp[current_weight] + item.value)
+    maximum_value = 0
```
ankit167: we could simply return **max(dp)** here

karpinski: Fair point! Done.

ankit167:
@karpinski get_maximum_value([Item(60, 10), Item(100, 20), Item(120, 30)],50)) gives output 300 instead of 220

karpinski:
@ankit167 My bad, it should be fine now.

PR #92, opened 2017-06-25
2-SAT algorithm

crismaz:
I implemented an algorithm for the 2-SAT problem.

It's a classic algorithm that constructs an implication graph, finds its strongly connected components, and assigns true/false values to them in topological order.


keon:
@crismaz Thanks for the PR.
sorry for the late response btw

PR #94, opened 2017-07-01
Add three solutions for kth_to_last and local tests

JacobIRR:
I initially put two files into the root folder, so I moved the changes into the linked list folder then deleted those in the root dir.


linkedlist/detect_loop.py
```diff
@@ -0,0 +1,50 @@
+def detect_loop(head):
```
ankit167: I believe, we can do this without extra space, using 'fast' and 'slow' pointers.

PR #95, opened 2017-07-02
add matrix transformations

ofek:


matrix/copy_transform.py
```diff
@@ -0,0 +1,62 @@
...
+            except IndexError:
+                new.insert(i, [])
+                new[i].append(elem)
+    return new
+
+def rotate_counterclockwise(matrix):
+    new = []
+    for row in matrix:
+        for i, elem in enumerate(reversed(row)):
+            try:
```
keon: why don't we use `defaultdict(list)` ?

ofek:
ready to merge?

matrix/copy_transform.py
keon: Oh i didn't know it was a list.
Ok but can we think of a way without using try and catch?
I'll accept the PR, but I am looking for a way to simplify this.

PR #100, opened 2017-09-24
Fix LinkedListQueue enqueue and dequeue

huhuta:


keon:
oh damn... you're right, thanks so much!

PR #102, opened 2017-10-26
Add different implementation of reverse_string.py

Akyshnik:
Add a simple and pythonic function of reversing a string.


keon:
on personal note, idk why `[::-1]` is pythonic.
pythonic code has to be explicit and I thought that is why they got rid of `++` and `--`.
Thanks anyways :)

PR #104, opened 2017-10-28
Added two graph algorithms

halloTheCoder:


keon:
Could you rename the filenames as snake cases?

PR #106, opened 2017-10-31
Added the egg dropping problem in the DP repo

ashaywalke:


keon:
Could you convert this to snake cases?

PR #110, opened 2017-10-31
Create min_cost_path.py

shivam207:
Find the minimum cost to reach destination using a train

### DIFF
--- a/dp/min_cost_path.py
+++ b/dp/min_cost_path.py
@@ -1,0 +1,27 @@
+# Python program to find min cost path 
+# from station 0 to station N-1
+ 
+global N
+N = 4
+def minCostRec(cost, s, d):
+ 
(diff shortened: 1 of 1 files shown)


dp/min_cost_path.py
```diff
@@ -0,0 +1,27 @@
+# Python program to find min cost path 
+# from station 0 to station N-1
+ 
+global N
+N = 4
+def minCostRec(cost, s, d):
```
keon: snake case

dp/min_cost_path.py
```diff
@@ -0,0 +1,27 @@
+# Python program to find min cost path 
+# from station 0 to station N-1
+ 
+global N
```
keon: no need global here

PR #112, opened 2017-11-02
Optimized circular_counter code, faster

mondeja:
If we store in memory len(list) we can run this algorithm too much faster.


array/circular_counter.py
```diff
@@ -14,9 +14,11 @@
 def josepheus(int_list, skip):
   skip = skip - 1 #list starts with 0 index
   idx = 0
-  while len(int_list)>0:
-    idx = (skip+idx)%len(int_list) #hashing to keep changing the index to every 3rd
+  len_list = (len(int_list))
+  while len_list>0:
```
ankit167: nit, goof to use spaces between operators.

PR #122, opened 2018-01-30
add README-CN.md 

yunshuipiao:
for Chinese students


keon:
my only concern is that if it can keep up to date with the main README.md 
I agree that having more language will improve the accessibility though.

README.md
```diff
@@ -1,3 +1,5 @@
+### English | [简体中文](https://github.com/yunshuipiao/algorithms/blob/master/README_CN.md)
```
keon: I think making it smaller would be better. (normal font size)
example: https://github.com/kdn251/interviews

yunshuipiao: @keon thank you .I will fix it .

yunshuipiao: @keon fix it . I will keep up to date with main README.md and the Chinese part as far as I can.

PR #127, opened 2018-02-08
Add Tarjan's strongly connected component finder

murilocamargos:
Hi @keon. maybe adding the Tarjan's strongly connected components finder algorithm would be a good addition.

As the name says, Tarjan's algorithm finds these type of components in a graph. These components are subgraphs where every vertex can be reached from any other vertex in the same component.

In social networks, this could indicate a close group of friends where everyone knows everyone. Another algorithm for this same task is the Kosaraju's algorithm.


keon:
sorry for the late reply.
Thanks @murilocamargos !

PR #129, opened 2018-02-24
update count_ones.py

goswami-rahul:
Complexity is better than the previous approach since number of loops is
   equal to the number of 1s in the binary representation.
Also (n < 0) check is not needed since it is specified unsigned integer.


goswami-rahul:
@keon Something wrong with this implementation?
If not, please merge this commit.
Also review and merge my some other PRs :)

keon:
Sorry about the delay. 
I'll try to review ASAP.

PR #130, opened 2018-02-24
refactored modules which shadow builtin module names

goswami-rahul:
q.qsize()>0  ---->   not q.empty()      #  is more verbose and pythonic ;)


heap/merge_sorted_k_lists.py
```diff
@@ -33,9 +33,9 @@ def merge_k_lists(lists):
     q = PriorityQueue()
     for node in lists:
         if node: q.put((node.val,node))
-    while q.qsize()>0:
-        curr.next = q.get()[1]
-        curr=curr.next
+    while not q.empty():
+        curr.next = q.get()[1]     # These two lines seem to 
+        curr=curr.next             # be equivalent to :-   curr = q.get()[1]
```
goswami-rahul: @keon please review the comment in
```python
curr.next = q.get()[1]
curr = curr.next
```
instead can we do just
```python
curr = q.get()[1]
```

keon:
the modules in this library are just to demonstrate, not to actually be used.
I think we could stick to the names without "my".

goswami-rahul:
@keon the builtin `queue` module is used in `heap/merge_sorted_k_lists.py`, and it throws error to use `PriorityQueue`. Also, it may be used in some further implementations.

goswami-rahul:
refer to #146

README.md
```diff
@@ -60,24 +60,33 @@ Minimal and clean example implementations of data structures and algorithms in P
...
     - [max_product_subarray](dp/max_product_subarray.py)
     - [max_subarray](dp/max_subarray.py)
+    - [min_cost_path](dp/min_cost_path.py)
     - [num_decodings](dp/num_decodings.py)
     - [regex_matching](dp/regex_matching.py)
+    - [rod_cut](dp/rod_cut.py)
     - [word_break](dp/word_break.py)
 - [graph](graph)
-    - [2-sat](graph/satisfiability.py)
+    - [strongly_connected](graph/checkDiGraphStronglyConnected.py)
```
keon: we never use camel case in this repo

PR #139, opened 2018-03-02
used split() to seperate words.

goswami-rahul:


goswami-rahul:
@SaadBenn The use of `string[::-1]` will reverse the string including all the characters in reverse order.
However, we need to reverse the sequence of words in the sentence. Therefore, the need of `arr = string.strip().split()`.
The `arr` contains the list of words separated.
We can do `arr = arr[::-1]` to reverse it.
But we choose the user defined method `reverse()` to show the use of algorithm.

PR #146, opened 2018-03-13
resolve conflict from refractoring modules. 

goswami-rahul:
resolving conflicts from #130 
@keon changed modules names 
- `math` ->  `maths`
- `string` -> `strings`
- `queue` -> `queues`


keon:
@goswami-rahul thanks for the clean up!

PR #157, opened 2018-03-25
linkedlist: added a testsuite and two functions for converting

christianbender:
I added two new functions:
```python
def convert_to_list(number : int) -> "Node":
def convert_to_str(l : "Node") -> str:
```
And I added a testsuite for the linked list and the full functionality of the project (uses unittest).

### DIFF
--- a/linkedlist/add_two_numbers.py
+++ b/linkedlist/add_two_numbers.py
@@ -19,7 +19,7 @@ class Node:
         self.next = None
 
 
-def add_two_numbers(left:"Node", right:"Node")->"Node":
+def add_two_numbers(left: Node, right: Node )-> Node:
     head = Node(0)
     current = head
     sum = 0
@@ -39,7 +39,7 @@ def add_two_numbers(left:"Node", right:"Node")->"Node":
     return head.next
 
 
-def convert_to_list(number : int) -> "Node":
+def convert_to_list(number : int) -> Node:
     """
         converts a positive integer into a (reversed) linked list.
         for example: give 112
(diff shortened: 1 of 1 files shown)


linkedlist/add_two_numbers.py
```diff
@@ -38,16 +39,91 @@ def add_two_numbers(left:"Node", right:"Node")->"Node":
     return head.next
 
 
+def convert_to_list(number : int) -> "Node":
+    """
```
goswami-rahul: The `"Node"` should be just `Node` in the typing hint.

linkedlist/add_two_numbers.py
```diff
@@ -38,16 +39,91 @@ def add_two_numbers(left:"Node", right:"Node")->"Node":
...
+            current = current.next
+            remainder = quotient % 10
+            quotient //= 10
+        current.next = Node(remainder)
+        return head.next
+    else:
+        print("number must be positive!")
+
+
+def convert_to_str(l : "Node") -> str:
```
goswami-rahul: Same applies here. `"Node"` change to `Node`.

goswami-rahul:
Good work. I have left a minor change in review.

christianbender:
@goswami-rahul  The orginal code contains "Node", too. That is why I wrote "Node". See here: https://github.com/christianbender/algorithms/blob/master/linkedlist/add_two_numbers.py
But I will change my code across your request.

goswami-rahul:
Oh I didn't see that. Thanks for the changes.

christianbender:
@goswami-rahul  No problem! Thanks for merging!

PR #161, opened 2018-03-26
Changed prime test.py

christianbender:
I removed a unnecessary 'else'. And I added a test suite (unittest).

### DIFF
--- a/maths/prime_test.py
+++ b/maths/prime_test.py
@@ -1,7 +1,7 @@
 """
 prime_test(n) returns a True if n is a prime number else it returns False
 """
-
+import unittest
 
 def prime_test(n):
     if n <= 1:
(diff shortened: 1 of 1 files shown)


keon:
@christianbender do you want to be a collaborator on this project?
Since you are building all test cases, you will eventually know every part of this project.
You are still recommended to post PR instead of directly committing since we all can review the code.

goswami-rahul:
@keon I think it is good idea to have @christianbender as a collaborator. It will help to have some active collaborators, since I think @ankit167 is not active these days.

christianbender:
@keon @goswami-rahul 
Yes, of course. I would like be a collaborator.

keon:
@christianbender Welcome!! 👍 I just sent the invitation (https://github.com/keon/algorithms/invitations)

PR #164, opened 2018-03-26
FIX flake8 errors, add test, refactor

JosipHaboic:
Fix flake8 errors.
Use python >3.5 syntax.
Do doc tests.
Some refactoring.
Try to add more info.

### DIFF
--- a/array/circular_counter.py
+++ b/array/circular_counter.py
@@ -1,4 +1,6 @@
 """
+https://en.wikipedia.org/wiki/Josephus_problem
+
 There are people sitting in a circular fashion,
 print every third member while removing them,
 the next counter starts immediately after the member is removed.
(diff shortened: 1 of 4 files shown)


array/circular_counter.py
```diff
@@ -7,42 +9,32 @@
...
+>>> josepheus_generator = josepheus(a, 3)
+>>> assert next(josepheus_generator) == '3'
+>>> assert next(josepheus_generator) == '6'
+>>> assert next(josepheus_generator) == '9'
+>>> assert next(josepheus_generator) == '4'
+>>> assert next(josepheus_generator) == '8'
+>>> assert next(josepheus_generator) == '5'
+>>> assert next(josepheus_generator) == '2'
+>>> assert next(josepheus_generator) == '7'
+>>> assert next(josepheus_generator) == '1'
```
goswami-rahul: `unittest` is preferred over `doctests`, as many algorithms already have unittests and it would be good to have consistency.

array/flatten.py
```diff
@@ -1,26 +1,36 @@
+from collections import Iterable
+from types import GeneratorType
```
goswami-rahul: `Iterable` is fine but I think the previous `collections.Iterable` was better here.  
The `GeneratorType` is not needed, `yield` clearly specifies a `generator`.

array/flatten.py
```diff
@@ -1,26 +1,36 @@
...
+>>> nested_list = [2, 1, [3, [4, 5], 6], 7, [8]]
+>>> flattened = flatten(nested_list)
+>>> assert next(flattened) == 2
+>>> assert next(flattened) == 1
+>>> assert next(flattened) == 3
+>>> assert next(flattened) == 4
+>>> assert next(flattened) == 5
+>>> assert next(flattened) == 6
+>>> assert next(flattened) == 7
+>>> assert next(flattened) == 8
```
goswami-rahul: `doctest` -> `unittest`

array/flatten.py
```diff
@@ -1,26 +1,36 @@
...
-import collections
 
 
-def flatten(inputArr, outputArr=None):
-    if not outputArr:
-        outputArr = []
-    for ele in inputArr:
-        if isinstance(ele, collections.Iterable):
-            flatten(ele, outputArr)
+def flatten(iterable: Iterable) -> GeneratorType:
```
goswami-rahul: Maybe this should be `flatten_v2` for generator version.

array/merge_intervals.py
```diff
@@ -4,51 +4,86 @@
...
-        else:
-            out.append(i)
-    return out
+
+    def __init__(self, start=0, end=0):
+        self.start = start
+        self.end = end
+
+    def __repr__(self):
+        return f"Interval [{self.start}, {self.end}]"
```
goswami-rahul: The `[` and `]` should be `(` and `)` to represent the constructor call.

array/flatten.py
goswami-rahul: The type `Iterable` is also unnecessary. The name itself is `iterable`!

JosipHaboic:
Cool thanks on review, will try to do it today.

PR #165, opened 2018-03-27
UPDATE add tests, refactor, fix flake8 errors

JosipHaboic:
Format, fix flake8 errors, refactor.
Would be cool when there would be template for examples,
something like this:
```python
"""
Algorithm description
"""

def my_algorithm():
    return "My new algorithm!"

if __name__ == "__main__":
    import unittest
    class TestMyAlgorithm(unittest.TestCase):
        """ Do the tests here. """
        pass
```

### DIFF
--- a/array/circular_counter.py
+++ b/array/circular_counter.py
@@ -19,22 +19,23 @@ def josepheus(int_list, skip):
         yield int_list.pop(idx)
         len_list -= 1
 
+import unittest
+
+class TestJosepheus(unittest.TestCase):
+    def test_josepheues(self):
+        a = ['1', '2', '3', '4', '5', '6', '7', '8', '9']
+        josepheus_generator = josepheus(a, 3)
+        self.assertEqual(next(josepheus_generator), '3')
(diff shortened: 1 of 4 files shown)


array/circular_counter.py
```diff
@@ -9,40 +9,32 @@
...
+    while len_list > 0:
+        idx = (skip+idx) % len_list  # hash index to every 3rd
+        yield int_list.pop(idx)
+        len_list -= 1
+
+
+if __name__ == "__main__":
+    import unittest
+
+    class TestJosepheus(unittest.TestCase):
```
goswami-rahul: the `import unittest` and testcases should be outside the `name == '__main__` block.
As an example, see how they are written [here](https://github.com/christianbender/algorithms/blob/372f3aaed2f789df34ea3641e592598ebe7869a8/linkedlist/add_two_numbers.py).

goswami-rahul:
@JosipHaboic yes it is preferred to have unittests in every algorithm, but the test cases are to be defined outside `name == __main__`. See my comment on the file.

PR #166, opened 2018-03-27
Improved and test suite for anagram.py

christianbender:
* I improved the function anagram (removed the while loop):
```python
def anagram(s1,s2):
```  
* I renamed one of the functions ```all_perms```  
```python
def all_perms_iter(elements):
```  
* And I added a test suite (unittest)

### DIFF
--- a/backtrack/anagram.py
+++ b/backtrack/anagram.py
@@ -41,7 +41,7 @@ def anagram(s1,s2):
     stillOK = True
     while j<26 and stillOK:
         if c1[j]==c2[j]:
-            j = j + 1
+            j += 1
         else:
             stillOK = False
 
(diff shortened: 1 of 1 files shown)


backtrack/anagram.py
```diff
@@ -31,14 +37,22 @@ def anagram(s1,s2):
...
-        if c1[j]==c2[j]:
-            j = j + 1
-        else:
-            stillOK = False
+    return c1 == c2
+
+
+class TestSuite (unittest.TestCase):
+    def test_all_perms(self):
+        allPerms = ['abc', 'bac', 'bca', 'acb', 'cab', 'cba']
```
goswami-rahul: The code should be formatted according to PEP guide.
Please use `snake_case` and avoid ` camelCase` for everything other than classes. Also there should be a space after each `,` and a newline before or after a function.

goswami-rahul:
There's a minor pep fix. Besides that, good work :+1:

christianbender:
@goswami-rahul  I made the required changes.

PR #167, opened 2018-03-27
Fix Python 3 related issues

cclauss:
Get us closer to compatibility with Python 3 without losing any existing Python 2 compatibility...
* __print()__ is a function in Python 3
* missing imports
* missing indentation
* define __raw_input()__, __reduce()__, etc.
* misc PEP8 cleanup

### DIFF
--- a/matrix/count_paths.py
+++ b/matrix/count_paths.py
@@ -1,5 +1,3 @@
-
-
 #
 # Count the number of unique paths from a[0][0] to a[m-1][n-1]
 # We are allowed to move either right or down from a cell in the matrix.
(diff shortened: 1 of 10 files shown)


matrix/count_paths.py
```diff
@@ -10,6 +8,13 @@
 #                           matrix. Return count[m-1][n-1]
 # T(n)- O(mn), S(n)- O(mn)
 #
+
+try:
+    raw_input          # Python 2
+except NameError:
+    raw_input = input  # Python 3
```
goswami-rahul: The repo is now Python 3+ so we can replace with `input`. No need for `raw_input`.

goswami-rahul:
Looks good!

matrix/count_paths.py
goswami-rahul: Infact, there is even no need of `input`. These are just implementations of algorithms. We can hardcode some inputs and add them in tests rather than taking input.

PR #170, opened 2018-03-27
add Christian as a maintainer

keon:
@christianbender :)


christianbender:
@keon  Thanks for adding!

PR #175, opened 2018-03-28
hashtable.py: Changed hashtable

christianbender:
* I added a runner:  
```python
if __name__ == "__main__":
    unittest.main()
```  
* I simplified the code.

### DIFF
--- a/map/hashtable.py
+++ b/map/hashtable.py
@@ -14,8 +14,8 @@ class HashTable(object):
     in Return True for a statement of the form key in map, if the given key is in the map, False otherwise.
     """
 
-    _empty = object()
-    _deleted = object()
+    _empty = None
+    _deleted = None
 
     def __init__(self, size=11):
         self.size = size
@@ -27,7 +27,7 @@ class HashTable(object):
         initial_hash = hash_ = self.hash(key)
 
         while True:
-            if self._keys[hash_] is self._empty or self._keys[hash_] is self._deleted:
+            if not self._keys[hash_]:
                 # can assign to hash_ index
                 self._keys[hash_] = key
                 self._values[hash_] = value
@@ -35,7 +35,6 @@ class HashTable(object):
                 return
             elif self._keys[hash_] == key:
                 # key already exists here, assign over
-                #self._keys[hash_] = key
                 self._values[hash_] = value
                 return
 
--- a/map/hashtable.py
+++ b/map/hashtable.py
@@ -47,7 +47,7 @@ class HashTable(object):
     def get(self, key):
         initial_hash = hash_ = self.hash(key)
         while True:
-            if self._keys[hash_] is self._empty:
+            if not self._keys[hash_]:
                 # That key was never assigned
                 return None
             elif self._keys[hash_] == key:
@@ -62,7 +62,7 @@ class HashTable(object):
     def del_(self, key):
         initial_hash = hash_ = self.hash(key)
         while True:
-            if self._keys[hash_] is self._empty:
+            if not self._keys[hash_]:
                 # That key was never assigned
                 return None
             elif self._keys[hash_] == key:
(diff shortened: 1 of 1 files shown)


map/hashtable.py
```diff
@@ -13,8 +14,8 @@ class HashTable(object):
     in Return True for a statement of the form key in map, if the given key is in the map, False otherwise.
     """
 
-    _empty = object()
-    _deleted = object()
+    _empty = None
+    _deleted = None
```
goswami-rahul: `_empty` and `_deleted` act as placeholders for not yet assigned keys and deleted keys respectively. 
In many places, there are uses like  `self.keys[hash_] is _empty` and `self.keys[hash_] is _deleted` which won't work as expected if be replace them both by `None`.

goswami-rahul:
I think the HashTable is designed such a way that you can only fill it with atmost `size` elements. If we delete a key from it, it is labeled as `_deleted` and no further key would be added to that hash. That means

map/hashtable.py
christianbender: @goswami-rahul  I ran all tests successful. After each change I ran the tests and it seems no problem.

christianbender:
@goswami-rahul 
I ran all tests successful. After each change I ran the tests and it seems no problem.   
I now changed the tests and ran it successful.  
I changed  
```python
def test_delete_key(self):
```
---

```python
def test_changes_on_hash_table(self):
        m = HashTable(10)
        for i in range(10):
            m.put(i,i**2)
        print("\t --> "+str(m.get(2))) # DEBUG
        for i in range(10):
            self.assertEqual(i**2,m.get(i))
```
The test (above) fails.  ```m.get(0) = None```  

**important:** I guess the hash-function is the problem. If we use a better hash-function the problem can be solve. 

At the moment:  
```python 
def hash(self, key):
        return key % self.size
```

map/hashtable.py
```diff
@@ -26,15 +27,14 @@ def put(self, key, value):
         initial_hash = hash_ = self.hash(key)
 
         while True:
-            if self._keys[hash_] is self._empty or self._keys[hash_] is self._deleted:
+            if not self._keys[hash_]:
```
goswami-rahul: I found the issue with `m.get(0) == None`. Here this line should be reverted to the original version or `if self._keys[hash_] is None:`

goswami-rahul:
`if not self._keys[hash_]:` confused also evaluated True for 0.

map/hashtable.py
```diff
@@ -47,7 +47,7 @@ def put(self, key, value):
     def get(self, key):
         initial_hash = hash_ = self.hash(key)
         while True:
-            if self._keys[hash_] is self._empty:
+            if not self._keys[hash_]:
```
goswami-rahul: revert back here too.

map/hashtable.py
```diff
@@ -62,7 +62,7 @@ def get(self, key):
     def del_(self, key):
         initial_hash = hash_ = self.hash(key)
         while True:
-            if self._keys[hash_] is self._empty:
+            if not self._keys[hash_]:
```
goswami-rahul: here

map/hashtable.py
```diff
@@ -118,7 +118,7 @@ def __resize(self):
         self._keys = [self._empty] * self.size
         self._values = [self._empty] * self.size
         for key, value in zip(keys, values):
-            if key is not self._empty and key is not self._deleted:
+            if key:
```
goswami-rahul: and here.

goswami-rahul:
I suggest that we keep the extra tests that you implemented and revert back `_empty` to `object()` as it makes more sense to have the placeholder of empty as a unique object, not even `None`. After this we can also make anything as the value in hashtable (including `None`).

christianbender:
@goswami-rahul  OK, I will see to it.

PR #178, opened 2018-03-31
completed tests for array module

goswami-rahul:
and some bugs fixes

### DIFF
--- a/array/longest_non_repeat.py
+++ b/array/longest_non_repeat.py
@@ -8,6 +8,7 @@ Given "pwwkew", the answer is "wke", with the length of 3.
 Note that the answer must be a substring,
 "pwke" is a subsequence and not a substring.
 """
+import unittest
 
 
 def longest_non_repeat(string):
@@ -45,7 +46,6 @@ def longest_non_repeat_two(string):
         used_char[char] = index
     return max_len
 
-import unittest
 
 class TestLongestNonRepeat(unittest.TestCase):
 
@@ -71,6 +71,7 @@ class TestLongestNonRepeat(unittest.TestCase):
         string = "pwwkew"
         self.assertEqual(longest_non_repeat_two(string), 3)
         
+
 if __name__ == "__main__":
     
     unittest.main()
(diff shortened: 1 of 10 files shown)


christianbender:
@goswami-rahul  I will view it tomorrow

array/longest_non_repeat.py
```diff
@@ -71,6 +71,7 @@ def test_longest_non_repeat_two(self):
         string = "pwwkew"
         self.assertEqual(longest_non_repeat_two(string), 3)
         
+
 if __name__ == "__main__":
```
christianbender: The empty line can be deleted.

christianbender:
The method print_intervals in merge_intervals.py can be simpler.

array/missing_ranges.py
```diff
@@ -1,27 +1,42 @@
...
 
-def missing_ranges(nums, lo, hi):
     res = []
     start = lo
-    for num in nums:
-        if num < start:
-            continue
-        if num == start:
+
+    for n in arr:
```
christianbender: Test your changes for the case:  
```arr := [3,4]```  
and 
``` lo := 1 ```  
``` hi := 5 ```  
The orginal code gets  
```['1->2', '5']```  
Your code should return ```[[1,2],[2,3],[3,5]]```

array/longest_non_repeat.py
goswami-rahul: Oh, my IDE puts it there automatically according to the PEP guide.

array/missing_ranges.py
goswami-rahul: Thanks for the review.
Yeah, I changed that to the `list of tuples` version, I feel denoting ranges in `tuples` is better than `str`.
And, for yours above example, the code should return `[(1, 3), (5, 5)]` (and it does).
The range basically means all numbers included in range `(l, r)` from `l` to `r`.  
I could have let `5` be just `5` rather than `(5, 5)`, but I think since the output is in form of tuples, `(5, 5)` is consistent with it.

christianbender: Ok, I made a fail.

goswami-rahul:
@christianbender Should I merge this?

goswami-rahul:
@christianbender Should I merge this?

christianbender:
@goswami-rahul Yes of course.

PR #184, opened 2018-04-02
Add tests to 'bit' module

goswami-rahul:
- added tests to files of `bit/` module
- wrote alternative method for `bit/find_missing_number.py`
- `bit/subset.py` now returns `Set[tuples]` rather than `List[lists]`.


bit/subsets.py
```diff
@@ -7,44 +7,54 @@
...
-def subnets(nums):
-    nums.sort()
-    total = 2 ** len(nums)  # or 1 << len(nums)
-    res = [] * total
+def subsets(nums):
+    """
+    :param nums: List[int]
+    :return: Set[tuple]
+    """
+    # nums.sort()  # no need for sort
```
christianbender: Can we don't remove this line? You commented it out.

README_CN.md
```diff
@@ -48,7 +48,7 @@ python版数据结构和算法实现的简约版小示例
     - [single_number2：寻找出现1次的数（2)](bit/single_number2.py)
     - [single_number:寻找出现1次的数（1)](bit/single_number.py)
     - [subsets: 求所有子集](bit/subsets.py)
-    - [add_without_operator：无操作符的加法](bit/add_without_operator.py)
+    - [add_bitwise_operator：无操作符的加法](bit/add_without_operator.py)
```
christianbender: What is with 无操作符的加法 ?

bit/count_ones.py
```diff
@@ -14,16 +14,58 @@
...
+
+def count_ones_recur(n):
     """Using Brian Kernighan’s Algorithm. (Recursive Approach)"""
-    if not n: return 0
-    return 1 + count_ones(n & (n-1))
 
-def count_ones(n):
+    if not n:
+        return 0
+    return 1 + count_ones_recur(n & (n-1))
```
christianbender: I think it must be ```return 1 + count_ones(n & (n-1))``` because you renamed the function.

README_CN.md
goswami-rahul: This is the Chinese version of README of this repo. Don't know if it is still maintained, maybe @keon can tell about this.

bit/count_ones.py
goswami-rahul: `return 1 + count_ones(n & (n-1))` was the old one. Now recursive version is renamed to `count_ones_recur`, and hence `return 1 + count_ones_recur(n & (n-1))`

bit/subsets.py
goswami-rahul: Yes, you are right. I will remove this.

bit/count_ones.py
christianbender: @goswami-rahul  OK.

christianbender:
@goswami-rahul I merge it.

PR #192, opened 2018-04-06
made the calculation of the mid simpler.

christianbender:


search/binary_search.py
```diff
@@ -9,7 +9,7 @@
 def binary_search(array, query):
     lo, hi = 0, len(array) - 1
     while lo <= hi:
-        mid = lo + (hi - lo) // 2
+        mid = (hi + lo) // 2
```
goswami-rahul: great, I think the previous one was implemented to avoid overflow in C-like languages, but there's no integer overflow in Python :)

PR #196, opened 2018-04-07
Changed sudoku solver

christianbender:
* I added a class-framework.  
* Added a simple test case with a ```main()``` method.
* Renamed the methods according to PEP.  
* Added a test for availability of ```xrange```  
* Added a ```__str__``` method. 

I will add a test suite (unittest)

### DIFF
--- a/dfs/sudoku_solver.py
+++ b/dfs/sudoku_solver.py
@@ -23,11 +23,6 @@ for the board lol).
 
 import unittest
 
-try:
-    xrange
-except:
-    xrange = range
-
 class Sudoku: 
     def __init__ (self, board, row, col):
         self.board = board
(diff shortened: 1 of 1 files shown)


dfs/sudoku_solver.py
```diff
@@ -21,59 +21,95 @@
...
-    return val
-
-def Solver(self):
-    if len(self.val)==0:
+import unittest
+
+try:
+    xrange
+except:
+    xrange = range
```
goswami-rahul: I don't think we should use `xrange` anymore. Please also change the usages of `xrange` to `range` for Python 3. Its also compatible with both versions 2 and 3.

christianbender:
@goswami-rahul I have done the requested changes.

PR #204, opened 2018-04-09
Implement bst tree

danghai:
Hi, I implement the bst tree. `insert`, and `search` works now. But I am not sure why `size` method does not work: 
The error:  `AttributeError: 'NoneType' object has no attribute 'size' `
@goswami-rahul  @christianbender Could you help me?

### DIFF
--- a/tree/bst/bst.py
+++ b/tree/bst/bst.py
@@ -1,0 +1,146 @@
+"""
+Implement Binary Search Tree. It has method:
+    1. Insert
+    2. Search
+    3. Size
+    4. Traversal (Preorder, Inorder, Postorder)
+"""
+
+import unittest
+
+class Node(object):
+    def __init__(self, data):
+        self.data = data
+        self.left = None
+        self.right = None
+
+class bst(object):
+    def __init__(self):
+        self.root = None
+
+    def get_root(self):
+        return self.root
+
+    """
+        Get the number of elements
+        Using recursion. Complexity O(logN)
+    """
+    def size(self):
+        return self.recur_size(self.root)
+
+    def recur_size(self, root):
+        if root is None:
+            return 0
+        elif root.left is None:
+            return 1 + self.recur_size(root.right)
+        elif root.right is None:
+            return 1 + self.recur_size(root.left)
+        else:
+            return 1 + self.recur_size(root.left) + self.recur_size(root.right)
+
+    """
+        Search data in bst
+        Using recursion. Complexity O(logN)
+    """
+    def search(self, data):
+        if self.root:
+            return self.recur_search(self.root, data)
+        else:
+            return False
+
+    def recur_search(self, root, data):
+        if root.data == data:
+            return True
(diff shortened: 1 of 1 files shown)


tree/bst/bst.py
```diff
@@ -0,0 +1,119 @@
...
+            return False
+        elif data > self.data:     # Go to right root
+            return self.right.search(data)
+        else:                      # Go to left root
+            return self.left.search(data)
+    """
+        Get the number of elements
+        Using recursion. Complexity O(logN)
+    """
+    def size(self):
```
goswami-rahul: these methods must be defined inside `BST` class, not inside `Node`.

danghai:
@tratneshsingh Thank you.

tree/bst/bst.py
danghai: What do you mean?

tratneshsingh:
For remove_max, think of a case where you remove the only node present in the tree. This shall give you a clue that you need to set the self.root = <>. So, should be the case for remove method.

danghai:
@tratneshsingh Hmm I do not get what you mean. I think I need to rewrite these method under BST first.

goswami-rahul:
@danghai I mean, class `Node` just represents a single node. It doesn't has any methods in itself. However, the methods like `size()` and `search()` does _not_ apply to _Node_, these apply to a _Tree_, such as our `BST`. Therefore, these entire methods should be moved to `BST` class.
Sorry, if you didn't understand what I mean, I will merge it and commit it myself :)

goswami-rahul:
Even better! I will make an abstract class `Tree` implementing these methods, and BST or any other tree can be subclassed from there.

danghai:
@goswami-rahul No I know what you mean :), I actually write these method under bst, However, I have struggle the node inside when I use recursion. I am going to rewrite them under bst now. 
Yeah, you help me to build them as a abstract class then

danghai:
@goswami-rahul Ok, I have updated. However, `remove_max` does not work.

tree/bst/bst.py
```diff
@@ -0,0 +1,189 @@
...
+        else:                      # Go to right root
+            if root.right:         # If right root is a node
+                return self.recur_insert(root.right, data)
+            else:
+                root.right = Node(data)
+                return True
+
+    def remove(self, data):
+        if self.root:
+            return self.root.remove(data)
```
goswami-rahul: There is no method as `remove` in class `Node`. Did you run your tests before committing?

goswami-rahul:
@danghai Please understand that incorrect code is not acceptable. If there are even some slight bugs, it is better to not include that method for now. They will surely be implemented by any of us in the future.

danghai:
@goswami-rahul , I do not implement the `remove` yet because the code `remove_max` does not work. I forget comment out

goswami-rahul:
Thanks. Looks good.

PR #206, opened 2018-04-09
Update first_occurance.py

tratneshsingh:
The earlier code fails to output the right index for following input: (Note that the first value '1' is repeated)
array = [1, 1, 2, 3, 3, 3, 3, 4, 4, 4, 4, 5, 6, 6, 6]
firstOccurance(array, 1)

### DIFF
--- a/search/first_occurance.py
+++ b/search/first_occurance.py
@@ -8,14 +8,16 @@ import unittest
 
 def firstOccurance(array, query):
     lo, hi = 0, len(array) - 1
-    while lo < hi:
+    while lo <= hi:
         mid = lo + (hi - lo) // 2
         #print("lo: ", lo, " hi: ", hi, " mid: ", mid)
+        if lo == hi:
+            break
         if array[mid] < query:
             lo = mid + 1
         else:
             hi = mid
-    if lo < len(array) and array[lo] == query:
+    if array[lo] == query:
       return lo
 
 class TestSuite(unittest.TestCase):
--- a/search/first_occurance.py
+++ b/search/first_occurance.py
@@ -4,42 +4,34 @@
 # T(n)- O(log n)
 #
 
+import unittest
 
 def firstOccurance(array, query):
-    lo, hi = 0, len(array) - 1
-    while lo <= hi:
-        mid = (hi + lo) // 2
-        print("lo: ", lo, " hi: ", hi, " mid: ", mid)
-        if (mid == 0 and array[mid] == query) or \
-           (array[mid] == query and array[mid-1] < query):
-            return mid
-        elif array[mid] <= query:
-            lo = mid + 1
-        else:
-            hi = mid - 1
+    lo, hi = 0, len(array)
+    while lo < hi:
+        mid = lo + (hi - lo) // 2
+        #print("lo: ", lo, " hi: ", hi, " mid: ", mid)
+        if array[mid] < query: lo = mid + 1
+        else: hi = mid
+    if lo < len(array) and array[lo] == query:
+      return lo
 
 
-def main():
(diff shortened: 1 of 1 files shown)


search/first_occurance.py
```diff
@@ -4,42 +4,34 @@
...
-    print("found first: ", query, " in index:", firstOccurance(array, query))
-    print("-----SEARCH-----")
-    query = 9
-    print("found first: ", query, " in index:", firstOccurance(array, query))
-
-if __name__ == "__main__":
-    main()
+    lo, hi = 0, len(array)
+    while lo < hi:
+        mid = lo + (hi - lo) // 2
```
christianbender: Here you can write ```(hi + lo) // 2```

christianbender: Write ```mid = (hi + lo) // 2```

search/first_occurance.py
```diff
@@ -4,42 +4,34 @@
...
-    print("-----SEARCH-----")
-    query = -1
-    print("found first: ", query, " in index:", firstOccurance(array, query))
-    print("-----SEARCH-----")
-    query = 9
-    print("found first: ", query, " in index:", firstOccurance(array, query))
-
-if __name__ == "__main__":
-    main()
+    lo, hi = 0, len(array)
```
christianbender: Instead of ```len(array)``` write ```len(array) -1```

search/first_occurance.py
```diff
@@ -4,42 +4,34 @@
...
+    while lo < hi:
+        mid = lo + (hi - lo) // 2
+        #print("lo: ", lo, " hi: ", hi, " mid: ", mid)
+        if array[mid] < query: lo = mid + 1
```
christianbender: Please write ```lo = mid +1``` in the next line. According to PEP8

search/first_occurance.py
```diff
@@ -4,42 +4,34 @@
...
+        else: hi = mid
```
christianbender: Same thing as above.

search/first_occurance.py
```diff
@@ -7,16 +7,19 @@
 import unittest
 
 def firstOccurance(array, query):
-    lo, hi = 0, len(array)
-    while lo < hi:
-        mid = lo + (hi - lo) // 2
+    lo, hi = 0, len(array) - 1
+    while lo <= hi:
+        mid = (hi + lo) // 2
```
keon: this can cause overflow.
the previous version was right.

search/first_occurance.py
```diff
@@ -7,16 +7,19 @@
...
         #print("lo: ", lo, " hi: ", hi, " mid: ", mid)
-        if array[mid] < query: lo = mid + 1
-        else: hi = mid
-    if lo < len(array) and array[lo] == query:
+        if lo == hi:
```
keon: I don't see how this is more preferable than the previous version.

search/first_occurance.py
tratneshsingh: Restored the previous one.

goswami-rahul: @keon I think there is no Integer overflow in Python. Or is there?

search/first_occurance.py
tratneshsingh: Done

search/first_occurance.py
tratneshsingh: Done

tratneshsingh: Done

search/first_occurance.py
tratneshsingh: Done

search/first_occurance.py
keon: @goswami-rahul you're right.
but still good to know for other langs tho.

PR #208, opened 2018-04-09
Add some problem to bst

danghai:
PR to add some problem to bst along with test case. A commit fix typo in `bst.py`

### DIFF
--- a/tree/bst/height.py
+++ b/tree/bst/height.py
@@ -1,0 +1,60 @@
+"""
+Write a function height returns the height of a tree. The height is defined to
+be the number of levels. The empty tree has height 0, a tree of one node has
+height 1, a root node with one or two leaves as children has height 2, and so on
+For example: height of tree is 4
+
+                    9
+                 /      \
+               6         12
+              / \       /   \
+            3     8   10      15
+                 /              \
+                7                18
+
+    height = 4
+
+"""
+import unittest
+from bst import Node
+from bst import bst
+
+def height(root):
+    if root is None:
+        return 0
+    else:
+        return 1 + max(height(root.left), height(root.right))
+
+"""
+    The tree is created for testing:
+
+                    9
+                 /      \
+               6         12
+              / \       /   \
+            3     8   10      15
+                 /              \
+                7                18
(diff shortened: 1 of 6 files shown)


tree/bst/height.py
```diff
@@ -0,0 +1,60 @@
...
+"""
+import unittest
+from bst import Node
+from bst import bst
+
+def height(root):
+    if root is None:
+        return 0
+    else:
+        return max(1 + height(root.left), 1 + height(root.right))
```
tratneshsingh: Please change this to the following:
return 1 + max(height(root.left), height(root.right))

tree/bst/num_empty.py
```diff
@@ -0,0 +1,69 @@
...
+
+"""
+import unittest
+from bst import Node
+from bst import bst
+
+def num_empty(root):
+    if root is None:
+        return 1
+    elif root.left is None and root.right is None:
```
tratneshsingh: Lines from 28-33 are redundant as the condition in 26 returns 1 if root is None.

danghai: Yeah, thanks

tree/bst/height.py
tratneshsingh: Try doing
``` python

# If you mean continuation, use the following.
return A + B + \
  C + D

# The below returns 'abc'
return reduce(operator.concat, ['a', 'b', 'c'])

# There are other ways to do the continuation : 'abc'
return ''.join(['a', 'b', 'c'])
```

goswami-rahul: @danghai Or you can simply do
```python
return (A + B
            + C + D)
```

tree/bst/num_empty.py
```diff
@@ -0,0 +1,67 @@
...
+
+"""
+import unittest
+from bst import Node
+from bst import bst
+
+def num_empty(root):
+    if root is None:
+        return 1
+    elif root.left is None and root.right:
```
tratneshsingh: As line 28 to line 31 are redundant, please remove

PR #217, opened 2018-04-11
Modify the flake8 option in travis

danghai:
`--max-line-length=<n>` option Set the maximum length that any line (with some exceptions) may be.
Exceptions include lines that are either strings or comments which are entirely URLs. 
The standard style coding max line length is 79.
We do not need to count number of errors too. We just need to know where exact error from the source. So I remove these option by `flake8 --show-source .`
You can refer the flake8 here:  http://flake8.pycqa.org/en/latest/user/options.html#

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -13,10 +13,7 @@ install:
     - pip install -r requirements.txt
     - pip install flake8  # pytest  # add another testing frameworks later
 before_script:
-    # stop the build if there are Python syntax errors or undefined names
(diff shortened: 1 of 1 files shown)


danghai:
It has a huge of errors, and needs the help to fix them. Or we can mention exact folder that we already fix in Travis ( Ex: `flake8 --show-source array`) until we already fill all error. And then we are able to use `.` instead.

danghai:
Reopen when team members consider `flake8` to test

Issue #162, opened 2018-03-26
separate test module


goswami-rahul:
Since we now have tests in most of the algorithms, I think it is a good idea to have a separate `test_module.py` file in each module. We can move all the tests from one module to it, and any further algorithms must have their tests in this file. It will make the repo easy to manage.
@keon what do you think? I can start working on this.

christianbender:
@goswami-rahul 
The problem is that we have not in all projects unit tests. Many projects (that I have see) have simple tests with the ```print``` function that prints something on the console. The idea with the central test-file sounds good.

goswami-rahul:
yeah those print statements need to be changed to tests too.  
But there is one disadvantage in moving tests to separate file, that the tests at the same place can also help viewers understand the algorithms better.

christianbender:
@goswami-rahul  I will jump in.

christianbender:
@goswami-rahul @keon 
I will test the module ```maths```.  I add a file ```test_module.py``` with all tests.

goswami-rahul:
@christianbender I would wait for starting work on this until @keon approves on this issue.

christianbender:
@goswami-rahul  OK, I will wait. Thanks for the notice

christianbender:
@danghai At the moment I pause on my work on this task. Because we don't have an approval.

keon:
@christianbender 

sorry for the late reply :(
As the repo gets larger, separating test code may be more scalable.
Let's give it a shot and see if it works better!

christianbender:
@keon @goswami-rahul #220

PR #223, opened 2018-04-13
Update general_solution.md

tratneshsingh:
Fix up the markup and add additional notes.

### DIFF
--- a/backtrack/general_solution.md
+++ b/backtrack/general_solution.md
@@ -1,156 +1,293 @@
-This structure might apply to many other backtracking questions, but here I am just going to demonstrate Subsets, Permutations, and Combination Sum.
 
-# Subsets : https://leetcode.com/problems/subsets/
+# Backtracking
+A backtracking algorithm tries to build a solution to a computational problem incrementally.Whenever the algorithm needs to decide between multiple alternatives to the next component of the solution, it simply tries all possible options recursively.
 
-public List<List<Integer>> subsets(int[] nums) {
-    List<List<Integer>> list = new ArrayList<>();
-    Arrays.sort(nums);
-    backtrack(list, new ArrayList<>(), nums, 0);
-    return list;
-}
+# General Pattern
+[Reference](http://jeffe.cs.illinois.edu/teaching/algorithms/notes/03-backtracking.pdf)
+1. Find a small choice whose correct answer would reduce the problem size.
+2. For each possible answer, temporarily adopt that choice and recurse. (Don’t try to be clever about which choices to try; just try them all.)
+3. The recursive subproblem is often more general than the original target problem; in each recursive subproblem, we must consider only solutions that are consistent with the choices we have already made
 
-private void backtrack(List<List<Integer>> list , List<Integer> tempList, int [] nums, int start){
-    list.add(new ArrayList<>(tempList));
-    for(int i = start; i < nums.length; i++){
-        tempList.add(nums[i]);
-        backtrack(list, tempList, nums, i + 1);
-        tempList.remove(tempList.size() - 1);
-    }
-}
-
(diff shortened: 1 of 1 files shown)


danghai:
Hi @tratneshsingh , could you squash 2 commits into 1? You can update the change commit by `git commit --amend`

tratneshsingh:
@danghai squashed multiple commits to one.

danghai:
@tratneshsingh , It has my commit in your PR. When you want to update the latest commit. 
Go back to master, and update the latest in your local master
```
git checkout master
git fech upstream
(If you do not have upstream: git remote add upstream <URL : upstream>)
git merge upstream/master
```
And then go to your branch. 
```
git checkout <your branch>
git rebase master
```
That is the way I do.

tratneshsingh:
Thans @hangdai

goswami-rahul:
I am not sure if this file is needed, cause it doesn't fit in this repo (being written in Java).

tratneshsingh:
I have converted the code in python and also it is good to add a
general_solution.md to each sub-directories if such a pattern exists.

backtrack/general_solution.md
```diff
@@ -1,156 +1,294 @@
...
+        helper(nums, target, idx+1, temp, csum+nums[idx], res)
+
+        # Unmark
+        temp.pop()
+```
+
+## Palindrome Partitioning
+https://leetcode.com/problems/palindrome-partitioning/
+```python
+def palindrome_partition(s):
```
goswami-rahul: I don't understand the need of these codes here again. Aren't these already implemented in the main algorithm files?

tratneshsingh:
Here, it is mentioned just to demonstrate common code pattern that is
applied to solve multiple different problems.

tratneshsingh:
But anyhow, if you intend to remove it, it is fine too. But after looking
at this file’s earlier version, it greatly helps in understanding other
code in the directory.

Thanks,
@tratneshsingh

SaadBenn:
@goswami-rahul @keon Do you guys still want to keep this file? I skimmed through the file and found that some of the algorithms are already implemented. It'd be redundant to have them both here. I was thinking that we should clean up a bit of the mess and keep the repo clean.

christianbender:
@SaadBenn @goswami-rahul @keon  
I would also suggest we remove the file. The code is also more Java.

tratneshsingh:
We should search for an solution approach in Python or pseudo code. .
>> This is what is done as part of this pull request. All java code
converted to python code with addition of generic pattern for pseudo code.

keon:
How about we split the non-redundant algos in a separate files.
and keep the explanation as README.md in this folder?

danghai:
Hi @tratneshsingh, are you interested in this issue #162 ?

keon:
Sure that was kind of my goal from the beginning but never got to that. 
My friend did some good works at https://github.com/kdn251/interviews

danghai:
@keon Before doing that, I think we should group all tests in a folder first. You can see in this PR #239

tratneshsingh:
I could create the test_stack.py for stack module.

Will pick others as we go along.

goswami-rahul:
So, I guess we can close this PR now, and delete the `general_solution.md`? #228

danghai:
@tratneshsingh I create test_stack.py in #245 . Could you do test_string.py? 
@keon I see the README begin longer following some files. I think we only put title (array, linkedlist, ...) in README and then writting all content file for each folder following (source, complexity, ...) However it requires a lot of work to do. If not, we simply list file name for each folder. What do you think?

tratneshsingh:
Sure would do that.

goswami-rahul:
Closing this PR.

PR #227, opened 2018-04-13
pep8 fixes and add random.seed()

goswami-rahul:


goswami-rahul:
Now, it seems to work fine.

maths/rsa.py
```diff
@@ -23,66 +23,72 @@
...
+    return b
+
+
+def generate_key(k, seed=None):
+    """
+    the RSA key generating algorithm
+    k is the number of bits in n
+    """
+
+    def gen_prime(k, seed=None):
```
christianbender: generates this function k-bits prime numbers?

maths/rsa.py
```diff
@@ -23,66 +23,72 @@
...
+
+        def is_prime(num):
+            if num == 2:
+                return True
+            for i in range(2, int(num ** 0.5) + 1):
+                if num % i == 0:
+                    return False
+            return True
+
+        random.seed(seed)
```
christianbender: Instead of ```random.seed(seed)``` you can simple write ```random.seed()``` . Then you can remove the unnecessary argumente of the function.

maths/rsa.py
goswami-rahul: Its the same function that was used previously, the code is short and seems correct to me.

maths/rsa.py
goswami-rahul: We have to provide a _seed_, because we need to create the similar random numbers for testing purposes. If we just call `random.seed()` without argument, it takes the system time as _seed_, which obviously varies, therefore I wrote `random.seed(seed)`.

goswami-rahul:
Fixed the merge conflicts. So, should we merge this?

Issue #226, opened 2018-04-13
bugs in math/rsa.py


goswami-rahul:
math/rsa.py sometimes shows up an error which means it still has some bugs. It passed previous tests because the data for rsa test is generated randomly, due which like it fails sometimes. This need to be fixed!
https://travis-ci.org/keon/algorithms/builds/366117956
when I restarted build, the test was passed.

christianbender:
@goswami-rahul 
I thought I had fixed the bug. I added ```random.seed()``` in the function ```def genprime(k):``` in the file rsa.py

christianbender:
@goswami-rahul 
I think the encryption level is to low with 8 bits.I wrote a new test with encryption level ```26 bits```. The test runs 100 times successful! If I minimize the encryption level the test fails! 

The original author used a encryption level of 1024 bits! On my machine a encryption level of 32 bits already  takes a while! 1024 bits for me impossible.

goswami-rahul:
@christianbender yes I changed some parameters of generate_key(). I added an optional argument `seed` which would help in testing. I also rewrote `is_prime` inside it, instead of importing from `rabin_miller` to resolve ambuiguity in #227 .

christianbender:
@goswami-rahul Can we close this issue?

PR #229, opened 2018-04-13
Fixed test_maths.py

christianbender:
**Fix for #226**  
* I changed the encryption level on 26 bits.  
* I wrote a new test which runs 100 times.  (successful).
  * With a high encryption level the encryption is save.  
     * disadvantage: The process takes an extremely long time.  
* I comment out this test, because it takes a while.

### DIFF
--- a/maths/rsa.py
+++ b/maths/rsa.py
@@ -22,7 +22,6 @@ a number very quickly:
 """
 
 import random
-import math
 
 from rabin_miller import * # is_prime
 from extended_gcd import * # extended_gcd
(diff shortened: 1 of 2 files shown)


goswami-rahul:
I think that importing `is_prime` from `rabin_miller` introduced that error, since it is not totally accurate. It checks for prime with a possibility of error. I rewrote is_prime in #227 . Please check it once.  
Once that is merged, we will merge this, it may lead to minor merge conflicts but we will resolve.

maths/rsa.py
```diff
@@ -22,10 +22,12 @@
 """
 
 import random
+import math
```
goswami-rahul: why this import?

christianbender: I removed it

PR #231, opened 2018-04-14
Refactor Tarjan

murilocamargos:
Hi guys, in this PR I'm refactoring my previous Tarjan's algorithm in order to keep up with the repo's new contributing guide. What I'm doing here is:
- Moving all the helper classes (Node, DirectedEdge and DirectedGraph) to graph.py file because they're general purpose classes that could be used by other graph algorithms;
- Making Tarjan algorithm a class to stop using global variables;
- Moving the examples to a test file so they could be used as test cases and serve as examples at the same time.

### DIFF
--- a/graph/graph.py
+++ b/graph/graph.py
@@ -82,10 +82,10 @@ class Graph:
         else:
             self.graph[u] = [v]
             
-g = Graph(4)
-g.add_edge(0, 1)
-g.add_edge(0, 2)
-g.add_edge(1, 2)
-g.add_edge(2, 0)
-g.add_edge(2, 3)
-g.add_edge(3, 3)
+#g = Graph(4)
+#g.add_edge(0, 1)
+#g.add_edge(0, 2)
+#g.add_edge(1, 2)
+#g.add_edge(2, 0)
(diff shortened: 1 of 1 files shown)


graph/graph.py
```diff
@@ -1,3 +1,69 @@
...
+            return node
+    
+    def add_edge(self, node_name_from, node_name_to):
+        try:
+            node_from = self.nodes[self.nodes.index(node_name_from)]
+            node_to = self.nodes[self.nodes.index(node_name_to)]
+            self.edges.append(DirectedEdge(node_from, node_to))
+        except ValueError:
+            pass
+
```
christianbender: Can you comment out the lines at line number 85
```python
g = Graph(4)
g.add_edge(0, 1)
g.add_edge(0, 2)
g.add_edge(1, 2)
g.add_edge(2, 0)
g.add_edge(2, 3)
g.add_edge(3, 3)
```

murilocamargos: They're actually not mine but it makes sense to comment them and create test cases for this later.

christianbender: Thanks.

PR #232, opened 2018-04-15
Add strip_url_params function.

SaadBenn:
The function removes duplicate query strings based off of an optional argument passed to the function. If the argument is empty then just remove any duplicates otherwise remove duplicates as well as the corresponding duplicates of the passed argument.

### DIFF
--- a/strings/strip_url_params.py
+++ b/strings/strip_url_params.py
@@ -64,16 +64,16 @@ def strip_url_params(url, params_to_strip=[]):
 
 # A very friendly pythonic solution (easy to follow)
 def strip_url_params(url, param_to_strip=[]):
-if '?' not in url:
-    return url
+    if '?' not in url:
+        return url
 
-queries = (url.split('?')[1]).split('&')
-queries_obj = [query[0] for query in queries]
-for i in range(len(queries_obj) - 1, 0, -1):
-    if queries_obj[i] in param_to_strip or queries_obj[i] in queries_obj[0:i]:
-        queries.pop(i)
+    queries = (url.split('?')[1]).split('&')
+    queries_obj = [query[0] for query in queries]
+    for i in range(len(queries_obj) - 1, 0, -1):
+        if queries_obj[i] in param_to_strip or queries_obj[i] in queries_obj[0:i]:
+            queries.pop(i)
 
-return url.split('?')[0] + '?' + '&'.join(queries)
+    return url.split('?')[0] + '?' + '&'.join(queries)
 
 
 # Here is my friend's solution using python's builtin libraries
--- a/strings/strip_url_params.py
+++ b/strings/strip_url_params.py
@@ -1,0 +1,106 @@
+"""
+Write a function that does the following:
+Removes any duplicate query string parameters from the url
+Removes any query string parameters specified within the 2nd argument (optional array)
+
+An example:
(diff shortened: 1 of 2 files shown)


SaadBenn:
Whoever is reviewing this file, please squash the commits into one when merging. When you click the `Merge pull request` button you'll see a squash option as well.

strings/strip_url_params.py
```diff
@@ -0,0 +1,106 @@
...
+Removes any duplicate query string parameters from the url
+Removes any query string parameters specified within the 2nd argument (optional array)
+
+An example:
+www.saadbenn.com?a=1&b=2&a=2') // returns 'www.saadbenn.com?a=1&b=2'
+"""
+
+# Here is a very non-pythonic grotesque solution
+from collections import defaultdict
+def strip_url_params(url, params_to_strip=[]):
```
goswami-rahul: please use `None` as default value for arguments in any function, rather than a mutable type (like `[]`). 
and since there are 3 algorithms, use names like `strip_url_params1`, `strip_url_params2`, etc.

strings/strip_url_params.py
```diff
@@ -0,0 +1,106 @@
...
+    queries_obj = [query[0] for query in queries]
+    for i in range(len(queries_obj) - 1, 0, -1):
+        if queries_obj[i] in param_to_strip or queries_obj[i] in queries_obj[0:i]:
+            queries.pop(i)
+
+    return url.split('?')[0] + '?' + '&'.join(queries)
+
+
+# Here is my friend's solution using python's builtin libraries
+import urlparse
```
goswami-rahul: library `urlparse` is not available anymore. It has been renamed as `urllib.parse`

strings/strip_url_params.py
```diff
@@ -0,0 +1,106 @@
...
+import urllib
+def strip_url_params(url, strip=None):
+    if not strip: strip = []
+    
+    parse = urlparse.urlparse(url)
+    query = urlparse.parse_qs(parse.query)
```
goswami-rahul: same changes. `urlparse` is renamed now.

strings/strip_url_params.py
```diff
@@ -0,0 +1,106 @@
...
+    
+    query = {k: v[0] for k, v in query.iteritems() if k not in strip}
+    query = urllib.urlencode(query)
```
goswami-rahul: `urllib.urlencode` been renamed to `urllib.parse.urlencode`.

strings/strip_url_params.py
```diff
@@ -0,0 +1,106 @@
...
+    return new.geturl()
+
+
+import unittest
+class TestSuite(unittest.TestCase):
+
+    def test_valid(self):
+        self.assertEqual(strip_url_params("www.saadbenn.com?a=1&b=2&a=2"), "www.saadbenn.com?a=1&b=2")
+        
+    def test_invalid(self):
```
goswami-rahul: We don't write tests for cases that are invalid! The tests will fail in such case, Travis will also show error.

goswami-rahul:
@SaadBenn `urllib` has some module renames in the new version
checkout [this](https://github.com/FriendCode/gittle/issues/49) and [this](https://stackoverflow.com/questions/28906859/module-has-no-attribute-urlencode) link.
Also see how the tests have been written in the file.

PR #234, opened 2018-04-16
Add algorithm that tests knowledge of strings

SaadBenn:


strings/validate_coordinates.py
```diff
@@ -0,0 +1,80 @@
...
+        longitude = float(l[1])
+    except:
+        return False
+    print(latitude, longitude)
+    return -90 <= latitude <= 90 and -180 <= longitude <= 180
+
+# friends solutions
+def is_valid_coordinates_1(coordinates):
+    try:
+        lat, lng = [abs(float(c)) for c in coordinates.split(',') if 'e' not in c]
```
christianbender: What means ```'e'``` ?

strings/validate_coordinates.py
```diff
@@ -0,0 +1,80 @@
...
+
+And some invalid ones:
+23.234, - 23.4234
+N23.43345, E32.6457
+6.325624, 43.34345.345
+0, 1,2
+
+"""
+# I'll be adding my attempt as well as my friend's solution (took us ~ 1 hour)
+
```
christianbender: Add ```import unittest```

strings/validate_coordinates.py
```diff
@@ -0,0 +1,80 @@
...
+	
+    def test_valid(self):
+    	valid_coordinates = [
+    		"-23, 25",
+    		"4, -3",
+    		"90, 180",
+    		"-90, -180"]
+
+    	for coordinate in valid_coordinates:
+    		self.assertTrue(is_valid_coordinates(coordinate))
```
christianbender: The function ```is_valid_coordinates(...)``` don't exists.

strings/validate_coordinates.py
```diff
@@ -0,0 +1,80 @@
...
+    def test_invalid(self):
+    	invalid_coordinates = [
+		"23.234, - 23.4234",
+		"99.234, 12.324",
+		"6.325624, 43.34345.345",
+		"0, 1,2",
+		"23.245, 1e1"]
+
+		for coordinate in invalid_coordinates:
+    		self.assertFalse(is_valid_coordinates(coordinate))
```
christianbender: The same thing as above.

strings/validate_coordinates.py
```diff
@@ -0,0 +1,80 @@
...
+def is_valid_coordinates_1(coordinates):
+    try:
+        lat, lng = [abs(float(c)) for c in coordinates.split(',') if 'e' not in c]
+    except ValueError:
+        return False
+
+    return lat <= 90 and lng <= 180
+
+# using regular expression
+import re
```
christianbender: Put this import to the begin of the file.

strings/validate_coordinates.py
```diff
@@ -0,0 +1,80 @@
...
+def is_valid_coordinates_regular_expression(coordinates):
+    return bool(re.match("-?(\d|[1-8]\d|90)\.?\d*, -?(\d|[1-9]\d|1[0-7]\d|180)\.?\d*$", coordinates))
```
christianbender: Please break up this line. max. 80 columns. if possible

strings/validate_coordinates.py
```diff
@@ -0,0 +1,68 @@
...
+N23.43345, E32.6457
+6.325624, 43.34345.345
+0, 1,2
+
+"""
+# I'll be adding my attempt as well as my friend's solution (took us ~ 1 hour)
+
+# my attempt
+def is_valid_coordinates_0(coordinates):
+    print(coordinates)
```
goswami-rahul: no need for `print` statements.

strings/validate_coordinates.py
goswami-rahul: I think he means to check for numbers with exponents like `12e5`. Or I may be completely wrong.

goswami-rahul:
@SaadBenn I see that you have created this branch in this original repo. That is OK, but it is preferred to maintain your changes in your own fork's branches, and then pull them here.

strings/validate_coordinates.py
SaadBenn: It checks for exponents.

strings/validate_coordinates.py
SaadBenn: I think It is always a good idea to limit the scope of import statements and variables and define them where they are needed. But If this is the requirement of this repo then I don't have problem changing it. @goswami-rahul

strings/validate_coordinates.py
SaadBenn: This is a regular expression. Breaking up a statement like this is very illogical, IMHO.

SaadBenn:
I'll heed to that in future. I'll delete this branch as soon as the merge is complete. @goswami-rahul

PR #235, opened 2018-04-17
Separate test for linkedlist, enhance and add some problem

### LINKED ISSUE #162: separate test module

danghai:
PR for separating test for linkedlist.
Fix #162 
I write a function convert from `Node` to `list` for testing. Now, we can use `assertEqual` in unittest. There are still some test case that need to be added. 
I am going to add some problem relating to them too. It is first draft, please review them.
@keon , @goswami-rahul

### DIFF
--- a/linkedlist/is_cyclic.py
+++ b/linkedlist/is_cyclic.py
@@ -4,15 +4,11 @@ Given a linked list, determine if it has a cycle in it.
 Follow up:
 Can you solve it without using extra space?
 """
-import unittest
-
-
 class Node:
 
     def __init__(self, x):
         self.val = x
         self.next = None
-
 
 def is_cyclic(head):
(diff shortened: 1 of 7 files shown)


christianbender:
@danghai  super! I merge this PR.

goswami-rahul:
@christianbender I think @danghai  was going to add more commits to this PR.

christianbender:
@danghai damn!

danghai:
@christianbender ,  It is fine. I create another PR then

PR #236, opened 2018-04-17
Add SeparateChainingHashTable implementation

mecforlove:
I just added a separate chaining hash table implementation here. Have a nice day!

### DIFF
--- a/map/separate_chaining_hashtable.py
+++ b/map/separate_chaining_hashtable.py
@@ -2,10 +2,6 @@ import unittest
 
 
 class Node(object):
-    key = None
-    value = None
-    next = None
-
     def __init__(self, key=None, value=None, next=None):
         self.key = key
         self.value = value
--- a/map/separate_chaining_hashtable.py
+++ b/map/separate_chaining_hashtable.py
@@ -1,5 +1,4 @@
 import unittest
-from unittest import TestCase
 
 
 class Node(object):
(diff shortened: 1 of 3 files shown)


map/separate_chaining_hashtable.py
```diff
@@ -0,0 +1,137 @@
...
+                if pre_node is None:
+                    self._table[hash_] = node_.next
+                else:
+                    pre_node.next = node_.next
+                self._len -= 1
+            pre_node = node_
+            node_ = node_.next
+
+    def hash(self, key):
+        return hash(key) % self.size
```
christianbender: Why you wrote this method recursively? Simple ```return key % self.size``` does the job, too.

map/separate_chaining_hashtable.py
```diff
@@ -0,0 +1,137 @@
...
+        self._len = 0
+        self._table = [self._empty] * size
+
+    def put(self, key, value):
+        hash_ = self.hash(key)
+        node_ = self._table[hash_]
+        if node_ is self._empty:
+            self._table[hash_] = Node(key, value)
+            self._len += 1
+            return
```
christianbender: Instead of the ```return``` you can write a ```else:``` and indent the code below.

map/separate_chaining_hashtable.py
```diff
@@ -0,0 +1,137 @@
+import unittest
+from unittest import TestCase
+
+
```
christianbender: Put in some comments in your code.

map/separate_chaining_hashtable.py
```diff
@@ -0,0 +1,137 @@
+import unittest
+from unittest import TestCase
```
christianbender: Remove this line. Write in your test class ```unittest.TestCase``` because you write in the main ```unittest.main()```

map/separate_chaining_hashtable.py
goswami-rahul: I think its not recursive. `hash()` is the builtin function here inside the defined method. Recursive would be `self.hash()`.

christianbender: @goswami-rahul Thanks for the note.

map/separate_chaining_hashtable.py
```diff
@@ -0,0 +1,147 @@
+import unittest
+
+
+class Node(object):
+    key = None
+    value = None
+    next = None
```
goswami-rahul: I think these attributes shouldn't be `class` attributes of `Node`. They should only be defined in `__init__`. Otherwise, only one copy gets initialized... or I maybe wrong..

mecforlove: yeah, I changed this.

map/separate_chaining_hashtable.py
mecforlove: I put comments in the class __doc__

Issue #29, opened 2017-04-26
Plan for testing


mingrammer:
I want to add some additional matrix algorithms and some tests for that. 

So, do you have a testing plan of test format, rules of in/output example data, test script name and others?

barrey11:
tell methe details.

mingrammer:
Well, I want to know following for writing test codes

How can I write the tests for some algorithms? Where? How? There are some rules? See example,
```python
# Algorithm logic

if __name__ == '__main__':
    test_case = ...
    assert algo(test_case) == expected_result    
```
or
```python
# In this case, how many test cases are needed?
# Algorithm logic

if __name__ == '__main__':
    test_cases = [...]
    for test in test_cases:
        assert algo(test_case) == expected_result    
```
or
```python
# This is test script named algo_test.py
import unittest
import algo

class SomeAlgoTest(...):
    def setUp(self):
        # set up code

    def tearDown(sef):
        # tear down code

    def test_some_algo(self):
        # declare some test cases and expected results
        # test logics
```

And how can I run the tests for all algorithms?

keon:
I think people should be able to run the algorithm using the implementation file. This is because we want people to learn and test algorithm with the minimalistic approach. But those simple tests cannot test the algorithm properly. So I suggest we put 4~5 simple tests in `__main__` and more extensive tests in the separate file. For now, we only have `__main__` (some don't even have it).

mingrammer:
@keon  Thanks for giving a suggestion. I'll use that way as following

```python
if __name__ == '__main__':
    test_cases = [
        {
            'input': <test_input_instance>,
            'expected': <expected_value_instance>,
        },
        ...
    ]

    for test in test_cases:
        assert algo(test['input']) == test['expected']
```

keon:
@mingrammer 
that's a good idea!

jprojs:
I think we must use assert function (TDD) with some tests for this purpose

Aluriak:
A good way to go could be to have dedicated `test_*.py` files in each directory.
It keeps the simplicity in source files, but allow one to expose more complete testing suite for an algorithm.
This last point is IMHO an important feature, mostly because it could provide a much more complete set of living examples, showing both general and corner cases to those who are interested in.

I'm a big fan of [`pytest`](http://pytest.org), which provides a very simple API (it just use the regular `assert` statement), and yet a complete output. Your testing code, @mingrammer, will not change, but will be moved into a dedicated function.
[...]

PR #239, opened 2018-04-20
Move all test into a folder

danghai:
PR to put all test into a folder `tests`. Enhance #29 
I think it is a great to put all test file into a folder named `tests`. 
Furthermore `pytest` in README does not work for me. It shows:
```
no test dir found testing here: /home/danghai/master/algorithms
*******************************************************************************
```
So, by putting all test file into a folder, I provide 2 ways to test in README instead. 
```
For running all tests write down:
    $ python3 -m unittest discover tests
For running some specific tests you can do this as following (Ex: sort):
    $ python3 -m unittest tests.test_sort
```
I already move all tests ( `linkedlist`, `sort`, `bit`) into tests folder. There are a lot of test for the rest of folder.  
What do you (@keon , @goswami-rahul , @christianbender , @SaadBenn ) think?


keon:
awesome! I like this direction.

danghai:
@goswami-rahul I think one of reason why I make the failure with `pytest` is PATH.  After command verson `pytest --version`: 
```
This is pytest version 3.5.0, imported from /usr/local/lib/python2.7/dist-packages/pytest.pyc
```

It should be inside python3.5. I am not sure how I can change the Path.

danghai:
@goswami-rahul I think one of solution for this issue is to use command : `python3 -m pytest tests`.
Could you try it with command `python3 -m pytest array` ( Because you do not have tests folder yet)?

danghai:
I am going to fix some others too

danghai:
@goswami-rahul  I think it is good to merge now ?

PR #245, opened 2018-04-21
Add heap and stack to test

danghai:
PR to add test_heap and test_stack to tests.

### DIFF
--- a/heap/binary_heap.py
+++ b/heap/binary_heap.py
@@ -29,7 +29,6 @@ For example remove_min() in a min heap:
 55 90 87             55  90                 55  90
 
 """
-import unittest
 from abc import ABCMeta, abstractmethod
 
 class AbstractHeap(metaclass=ABCMeta):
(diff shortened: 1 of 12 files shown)


heap/binary_heap.py
```diff
@@ -51,7 +50,7 @@ def min_child(self,i):
     @abstractmethod
     def remove_min(self,i):
         pass
-class Binary_Heap(AbstractHeap):
+class binary_heap(AbstractHeap):
```
goswami-rahul: This should be renamed to `BinaryHeap`.

stack/stack.py
```diff
@@ -49,7 +46,7 @@ def peek(self):
         pass
 
 
-class ArrayStack(AbstractStack):
+class Array_stack(AbstractStack):
```
goswami-rahul: I think it was correct before? `ArrayStack`.

stack/stack.py
```diff
@@ -101,7 +98,7 @@ def __init__(self, value):
         self.next = None
 
 
-class LinkedListStack(AbstractStack):
+class Linkedlist_stack(AbstractStack):
```
goswami-rahul: `LinkedListStack`

tests/test_heap.py
```diff
@@ -0,0 +1,52 @@
+from heap.binary_heap import binary_heap
+from heap.skyline import get_skyline
+from heap.sliding_window_max import max_sliding_window
+
+import unittest
+
+class Test_binary_heap(unittest.TestCase):
```
goswami-rahul: rename to `TestBinaryHeap`

tests/test_stack.py
```diff
@@ -0,0 +1,133 @@
...
+        self.assertFalse(is_valid("[[[]]"))
+        self.assertTrue(is_valid("{([])}"))
+        self.assertFalse(is_valid("(}"))
+
+    def test_simplify_path(self):
+        p = '/my/name/is/..//keon'
+        self.assertEqual('/my/name/keon', simplify_path(p))
+
+
+class Test_Stack(unittest.TestCase):
```
goswami-rahul: Why another class for it?

danghai: Because I think stack.py is a class with many methods. Other files is just function. I think it is more clear :)

christianbender: @danghai Please use ```TitleCase``` for class names.

danghai:
I have updated it!

PR #249, opened 2018-04-22
fix the error in is_palindrome

goswami-rahul:
fix the minor syntax error (indentation) in `strings/is_palindrome.py`


SaadBenn:
@goswami-rahul I added the string module at the same time you did.

goswami-rahul:
@SaadBenn we should merge this now.

Issue #247, opened 2018-04-22
Tests in directory tests 


christianbender:
[tests](https://github.com/keon/algorithms/tree/master/tests)  
If I ran the tests I get this kind of error message:  

> from search.binary_search import binary_search, binary_search_recur
ImportError: No module named 'search'

or   

> from array.delete_nth import delete_nth, delete_nth_naive
ImportError: No module named 'array.delete_nth'; 'array' is not a package

goswami-rahul:
@christianbender I think there is no `__init__.py` file in `array` and `search` module yet, that is why it is not recognized as a package.

danghai:
Hello @christianbender , could you give exact the command makes these error. 
btw, you miss the `__init__` in maths

danghai:
Thanks, that is all I am looking for

goswami-rahul:
@christianbender are you sure your branch is currently updated with the original repo?
Because your master branch is currently 220 commits behind this repo. Please update your fork, and try again if this error still shows up.

christianbender:
@goswami-rahul  My local branch on my machine is up to date. 
@danghai 
 > Hello @christianbender , could you give exact the command makes these error.
btw, you miss the __init__ in maths

I tried to ran any file. But this error (above) occures. In the example above I tried to ran the file ```test_search.py``` and the file ```test_array.py```.

danghai:
@christianbender  what is your command line?

christianbender:
@danghai @goswami-rahul  I'm a idiot! :angry: I started the files seperated ... :flushed: Now with the command ```python3 -m unittest discover tests``` it works perfect! :smirk:

PR #251, opened 2018-04-22
Fix test

danghai:
PR for fixing test #247


tests/test_array.py
```diff
@@ -1,16 +1,16 @@
...
-from array.longest_non_repeat import longest_non_repeat, longest_non_repeat_two
-from array.merge_intervals import Interval, merge_v2
-from array.missing_ranges import missing_ranges
-from array.move_zeros_to_end import move_zeros
-from array.plus_one import plus_one, plus_one_v2, plus_one_v3
-from array.rotate_array import rotate_v1, rotate_v2, rotate_v3
-from array.summary_ranges import summary_ranges
-from array.three_sum import three_sum
-from array.two_sum import two_sum
+from Array.delete_nth import delete_nth, delete_nth_naive
```
goswami-rahul: a better module name would be `arrays` for it.

tests/test_search.py
```diff
@@ -1,11 +1,11 @@
-from search.binary_search import binary_search, binary_search_recur
-from search.first_occurance import first_occurance
-from search.last_occurance import last_occurance
-from search.search_insert import search_insert
-from search.two_sum import two_sum, two_sum1, two_sum2
-from search.search_range import search_range
-from search.find_min_rotate import find_min_rotate, find_min_rotate_recur
-from search.search_rotate import search_rotate, search_rotate_recur
+from Search.binary_search import binary_search, binary_search_recur
```
goswami-rahul: no need to rename `search`, no built-in name `search`

goswami-rahul:
@danghai this issue was faced before also. I fixed it for some other modules in #146 . So, it would be better if we follow the same convention and rename `array` -> `arrays`. I didn't know about `array` module at time, that's why we are facing this name conflict now. see #146 .
and no need for `search` (I don't think there's a module named search??)

danghai:
I do not see any problem for `search` too, but #247 reports `search`

danghai:
Okay , I will update and merge it

Issue #240, opened 2018-04-21
Error: test case SeparateChainingHashTable implementation 


danghai:
I have error when I try to run `test_graph` for `tarjan`. I am not sure about this algorithm. 
Hello, @mecforlove . It is very nice when you take a review about it? Thanks
```
FF
======================================================================
FAIL: test_tarjan_example_1 (test_graph.TestTarjan)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/home/danghai/algorithms/graph/test_graph.py", line 28, in test_tarjan_example_1
    self.assertEqual(g.sccs, [['F', 'G'], ['H', 'D', 'C'], ['E', 'B', 'A']])
AssertionError: Lists differ: [[F, G], [H, D, C], [B, A, E]] != [['F', 'G'], ['H', 'D', 'C'], ['E', 'B', 'A']]

First differing element 2:
[B, A, E]
['E', 'B', 'A']

- [[F, G], [H, D, C], [B, A, E]]
+ [['F', 'G'], ['H', 'D', 'C'], ['E', 'B', 'A']]

======================================================================
FAIL: test_tarjan_example_2 (test_graph.TestTarjan)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/home/danghai/algorithms/graph/test_graph.py", line 44, in test_tarjan_example_2
    self.assertEqual(g.sccs, [['B', 'E', 'A'], ['D', 'C'], ['G', 'F'], ['H']])
AssertionError: Lists differ: [[E, A, B], [D, C], [F, G], [H]] != [['B', 'E', 'A'], ['D', 'C'], ['G', 'F'], ['H']]

First differing element 0:
[E, A, B]
['B', 'E', 'A']

- [[E, A, B], [D, C], [F, G], [H]]
+ [['B', 'E', 'A'], ['D', 'C'], ['G', 'F'], ['H']]

----------------------------------------------------------------------
Ran 2 tests in 0.002s

FAILED (failures=2)
```

mecforlove:
Can you provide more details? Is this related to `SeparateChainingHashTable`?

danghai:
@mecforlove I just look your commit in PR. 
It is fail when I run command to test the algorithm: `python3 -m unittest test_graph.py`. I think it has something wrong in algorithm. In particular: in test_graph
```
        g = Tarjan(example)
        self.assertEqual(g.sccs, [['F', 'G'], ['H', 'D', 'C'], ['E', 'B', 'A']])

[...]

PR #252, opened 2018-04-23
Fix tarjan tests

murilocamargos:
Hi @danghai and @keon, I've fixed the possible errors reported at #240 by sorting each strongly connected component. The sort is done by the Node's name, that's why I added comparison methods on Node class. I also moved the graph tests file to the new correct directory.

It should be working now!


danghai:
Thanks @murilocamargos. It is really awesome!

christianbender:
@goswami-rahul @danghai  I have merge this PR.

Issue #241, opened 2018-04-21
Bit Manipulation: Some interesting questions


danghai:
1) **Flip Bit to win:** You have an integer and you can flip exactly one bit from a `0` to `1`. Write code to find the length of the longest sequence of `1s` you could create. 
For example:
Input: 1775 ( or: `11011101111`)
Output: 8 

2) **Conversion:** Write a function to determine the number of bits you would need to flip to convert integer A to integer B.
For example:
Input:  29 (or: `11101`), 15 (or: `01111`)
Output: 2

goswami-rahul:
Not sure about the first problem, but the second problem can be solved by first taking the `xor` of the two numbers, and then count the number of set bits in it.

SaadBenn:
Those are very interesting questions. @danghai 
Bit manipulation is something that I lack.

danghai:
@SaadBenn I create some questions in bit manipulation in PR #253 (It does not contain 2 questions in this issue). You can contribute by adding more algorithm to solve the problem, or get more bit manipulation question.

SaadBenn:
@danghai Thanks. I'll get back to adding more algorithms but at the moment I've some big assignments going on.

PR #253, opened 2018-04-23
Add some bit manipulation

danghai:
PR for adding some bit manipulation. 
Relate to #241


bit/has_alternative_bit.py
```diff
@@ -0,0 +1,31 @@
...
+Output: False because the binary representation of 7 is: 111.
+
+Input: 11
+Output: False because the binary representation of 11 is: 1011.
+
+Input: 10
+Output: True because The binary representation of 10 is: 1010.
+"""
+
+def has_alternative_bit(n):
```
goswami-rahul: This can also be done using bit mask in O(1). Should I commit that algo in this same PR?

goswami-rahul:
@danghai check out another method I added.

danghai:
@goswami-rahul Oh it is better my solution with O(1)

bit/has_alternative_bit.py
danghai: I merge your commit in my RP. Thanks

danghai:
It is good to merge now.

PR #260, opened 2018-04-27
add single_number3: find two elements that appear only once

keon:
Interesting problem I solved today :)
Single number series is actually really good for practicing bitwise operations, especially XOR.


danghai:
It is interesting question! I come up with another problem being harder. Find the two elements that appear only one. All the other elements may appear twice, thrice, ... ^^

SaadBenn:
Can we not do this with a dict? Loop through the list and add the count. Then loop again and return the result whose count is 1. Time complexity would be `O(2n)` == `O(n)` @keon

keon:
@SaadBenn 
That would be the best answer if there is no constraint.
but in this case using dict will make the space complexity O(N). we want O(1)

keon:
@danghai we would need a different approach then because XOR wouldn't work for numbers appearing odd number of times.

bit/single_number3.py
```diff
@@ -0,0 +1,49 @@
...
+    """
+    :type nums: List[int]
+    :rtype: List[int]
+    """
+    # isolate a+b from pairs using XOR
+    ab = 0
+    for n in nums:
+        ab ^= n
+
+    # isolate right most bit from a+b
```
goswami-rahul: @keon Shouldn't this be `a^b` ??

bit/single_number3.py
```diff
@@ -0,0 +1,49 @@
...
+    right_most = ab & (-ab)
+
+    # isolate a and b from a+b
```
goswami-rahul: here too?

keon: thanks, done!

bit/single_number3.py
keon: thanks, done!

PR #273, opened 2018-05-03
Update priority_queue.py

goswami-rahul:
- refactor `insert` and `delete` -> `push` and `pop`. Because `pop` seems more natural while returning the deleted item (rather than `delete`)
- `push` now takes data and priority(optional), and constructs a `PriorityQueueNode` inside the method for inserting. This simplifies the usage.
- used `collections.deque()` for O(1) popleft operation.
- add `__repr__` method.
(refer #271 )

### DIFF
--- a/queues/priority_queue.py
+++ b/queues/priority_queue.py
@@ -1,6 +1,10 @@
 """
-Implementation of priority queue
+Implementation of priority queue using linear array.
+Insertion - O(n)
+Extract min/max Node - O(1)
 """
+import collections
+
 
 class PriorityQueueNode:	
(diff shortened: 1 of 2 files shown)


SaadBenn:
Excellent job @goswami-rahul I'll take a look at it.

SaadBenn:
Your `push` method is so much better than mine and overall the code is more compact. I like it. Great job at refactoring. @goswami-rahul  I think it's good to merge now but I'll wait for @danghai.

goswami-rahul:
@danghai Thanks for the tests :)

PR #285, opened 2018-05-23
Fix sort

techkang:
I added doc sting to wiggle_sort.py, modified counting_sort.py and merge_sort.py so that the type of returned array is also numpy.ndarray when the type of input array is numpy.ndarray (the other sort method already achieved this). I alos wrap function in quick_sort.py, so we can use quick sort just like other sort method. And I also modified all the files according to pep8 specification.

### DIFF
--- a/sort/comb_sort.py
+++ b/sort/comb_sort.py
@@ -9,7 +9,6 @@ from math import floor
 
 
 def comb_sort(arr):
-
     def swap(i, j):
         arr[i], arr[j] = arr[j], arr[i]
 
@@ -18,7 +17,7 @@ def comb_sort(arr):
     shrink = 1.3
     sorted = False
     while not sorted:
-        gap = int(floor(gap/shrink))
+        gap = int(floor(gap / shrink))
         if gap > 1:
             sorted = False
         else:
--- a/sort/comb_sort.py
+++ b/sort/comb_sort.py
@@ -5,7 +5,6 @@ https://en.wikipedia.org/wiki/Comb_sort
 Worst-case performance: O(N^2)
 
 """
-from math import floor
 
 
 def comb_sort(arr):
@@ -17,7 +16,7 @@ def comb_sort(arr):
     shrink = 1.3
     sorted = False
     while not sorted:
-        gap = int(floor(gap / shrink))
+        gap = int(gap / shrink)
         if gap > 1:
             sorted = False
         else:
--- a/sort/heap_sort.py
+++ b/sort/heap_sort.py
@@ -14,7 +14,7 @@ def max_heap_sort(arr):
 def max_heapify(arr, end):
     """ Max heapify helper for max_heap_sort
     """
-    last_parent = int((end - 1) / 2)
+    last_parent = (end - 1) // 2
 
     # Iterate from last parent to first
     for parent in range(last_parent, -1, -1):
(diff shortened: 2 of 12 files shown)


techkang:
check failed because I wraped the qucik sort funciton.

sort/heap_sort.py
```diff
@@ -2,28 +2,29 @@ def max_heap_sort(arr):
...
         arr[0] = arr[i]
         arr[i] = temp
     return arr
 
+
 def max_heapify(arr, end):
     """ Max heapify helper for max_heap_sort
     """
-    last_parent = int((end-1)/2)
+    last_parent = int((end - 1) / 2)
```
goswami-rahul: `(end - 1) // 2`

sort/comb_sort.py
```diff
@@ -18,7 +17,7 @@ def swap(i, j):
     shrink = 1.3
     sorted = False
     while not sorted:
-        gap = int(floor(gap/shrink))
+        gap = int(floor(gap / shrink))
```
goswami-rahul: This can be just `gap = gap // shrink`.

sort/heap_sort.py
```diff
@@ -42,34 +43,36 @@ def min_heap_sort(arr):
...
 def min_heapify(arr, start):
     """ Min heapify helper for min_heap_sort
     """
     # Offset last_parent by the start (last_parent calculated as if start index was 0)
-    # All array accesses need to be offet by start
-    end = len(arr)-1
-    last_parent = int((end-start-1)/2)
+    # All array accesses need to be offset by start
+    end = len(arr) - 1
+    last_parent = int((end - start - 1) / 2)
```
goswami-rahul: same. please change this with floor division `//`.

techkang:
@goswami-rahul Thanks, I changed code according to your advise.

sort/comb_sort.py
```diff
@@ -18,7 +16,7 @@ def swap(i, j):
     shrink = 1.3
     sorted = False
     while not sorted:
-        gap = int(floor(gap/shrink))
+        gap = int(gap / shrink)
```
goswami-rahul: Thanks for the edits. 
I think you missed this line. This line could be `gap //shrink`.

goswami-rahul:
@techkang Please also remove the extra 2 arguments from the tests of quicksort, so that the tests will pass.

sort/comb_sort.py
techkang: Because shrink is not a int, so gap // shrink is not a int neither.

techkang:
@goswami-rahul I added two default arguments to quick_sort to pass test. However, it seems that it doesn't work well. Can you help me out?

danghai:
Hi @techkang , I see it works. Could you elaborate it why it does not work?

sort/comb_sort.py
goswami-rahul: oh I see. Thanks for pointing out.

sort/quick_sort.py
```diff
@@ -1,17 +1,25 @@
-def quick_sort(arr, first, last):
-    """ Quicksort
+def quick_sort(arr, first=0, last=-999):
```
goswami-rahul: This looks fine. But the version with the single argument `arr` was better. I meant to say that you make changes in the test cases in tests/test_sort.py, and remove those extra arguments from the quicksort call.

goswami-rahul:
@techkang there is no issue with the tests. Travis build passed. I don't know why Coveralls is showing decreased coverage. Maybe its because of the newly added `quicksort_recur` function. Don't worry about it :)

goswami-rahul:
sorry i clicked close button by mistake. I am on my phone .

danghai:
@goswami-rahul If the Coverall decrease, it is okay. Because you may add more test case, or you change something in flow code ( if condition, call other functions, ...)

techkang:
@goswami-rahul I updated code and delete two arguments of quick sort.

PR #301, opened 2018-06-03
Created jump_seach in algorithms/seach

hsi1032:
I upload jump_sort and add it to test_search file


algorithms/search/jump_search.py
```diff
@@ -0,0 +1,38 @@
...
+    block_prev = 0
+    block= block_size
+
+    #return -1 means that array doesn't contain taget value
+    #find block that contains target value
+    while arr[min(block,n)-1] < target :
+        block_prev = block
+        block += block_size
+        if block_prev >=n :
+            return -1
```
goswami-rahul: The index of `arr` in the `while` condition looks cluttered, here, a better piece of code can be this. ( i hope it doesn't has some bug, feel free to point-out)
```python
if arr[n - 1] > target:
    return -1  # array doesn't contain target value
while block <= n and arr[block - 1] < target:
    block_prev = block
    block += block_size
```

algorithms/search/jump_search.py
```diff
@@ -0,0 +1,38 @@
+import math
+
+def jump_search(arr,target):
+    '''Jump Search
```
goswami-rahul: (pep8) docstrings should be encloses by triple _double_ quotes (`""" ... """`).

algorithms/search/jump_search.py
```diff
@@ -0,0 +1,38 @@
...
+        reference: https://en.wikipedia.org/wiki/Jump_search
+
+    '''
+    n = len(arr)
+    block_size = int(math.sqrt(n))
+    block_prev = 0
+    block= block_size
+
+    #return -1 means that array doesn't contain taget value
+    #find block that contains target value
```
goswami-rahul: please leave a space after `#` in the comments.

algorithms/search/jump_search.py
```diff
@@ -0,0 +1,38 @@
...
+    while arr[min(block,n)-1] < target :
+        block_prev = block
+        block += block_size
+        if block_prev >=n :
+            return -1
+
+    #find target value in block
+    while arr[block_prev] < target :
+        block_prev += 1
+        if block_prev == min(block,n):
```
goswami-rahul: there should be a space after `, ` in `min(block, n)`.

tests/test_search.py
```diff
@@ -88,6 +89,13 @@ def test_search_rotate(self):
         self.assertEqual(8, search_rotate_recur(array, 0, 11, 5))
         self.assertEqual(-1, search_rotate_recur(array, 0, 11, 9))
 
+    def test_jump_search(self):
+        array = [1, 2, 3, 3, 3, 3, 4, 4, 4, 4, 5, 6]
+        self.assertEqual(10, jump_search(array, 5))
+        self.assertEqual(11, jump_search(array, 6))
+        self.assertEqual(-1, jump_search(array, 7))
+        self.assertEqual(-1, jump_search(array, -1))
```
goswami-rahul: What does `jump_search(array, 3)` return here? it can be any of the `2, 3, 4, 5`.
I think it depends on the block_size (and hence array size), in the case of repeated elements.

Hence, this algorithm should only be used when we are fine with any index in case of duplicate elements. I think we should include a small note about this too in the docstring.

PR #302, opened 2018-06-03
Added link to Korean README file in other README files

9967han:
It seems like Korean README file is created, so I added link to Korean file from other README files - English, Chinese, German and Japanese.


danghai:
@9967han Thanks PR.

PR #303, opened 2018-06-04
Created radix_sort.py in algorithms/sort

aig031:
It seems like there are no radix_sort in the sort file, so I added it.

### DIFF
--- a/algorithms/sort/radix_sort.py
+++ b/algorithms/sort/radix_sort.py
@@ -1,0 +1,25 @@
+'''
+radix sort
+complexity: O(nk) . n is the size of input list and k is the digit length of the number
+'''
+def radix_sort(arr):
+    is_done = False
+    position = 1
+
+    while not is_done:
+        queue_list = [list() for _ in range(10)]
+        is_done = True
+
+        for num in arr:
+            digit_number = (int)(num/position)%10
+            queue_list[digit_number].append(num)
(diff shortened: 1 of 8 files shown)


aig031:
I change test_sort too!

algorithms/sort/radix_sort.py
```diff
@@ -0,0 +1,26 @@
...
+def radix_sort(arr):
+    is_done = False
+    position = 1
+
+    while not is_done:
+        queue_list = [list() for _ in range(10)]
+        is_done = True
+
+        for num in arr:
+            digit_number = (int)(num/position)%10
```
goswami-rahul: @aig031 Typecasting in Python is different from the C version (that you used here).
The correct syntax for typecasting is 
```python
digit_number = int(num / position) % 10
```
But the recommended way for this in Python _3_ is
```python
digit_number = num // position % 10
```
Also notice the spaces around the operators, for readability.

algorithms/sort/radix_sort.py
```diff
@@ -0,0 +1,26 @@
+'''
```
goswami-rahul: (pep8) docstrings should be encloses by triple _double_ quotes (`""" ... """`).

aig031:
Thanks for your comment, I updated radix_sort and, update README files.

PR #304, opened 2018-06-04
Created combination.py in maths

lg970325:
I created combination.py that calculates nCr. Also added test cases as well.


lg970325:
README work complete :)

PR #306, opened 2018-06-05
clean up backtrack and move tests to test_backtest

keon:
cleanin up.
just changed names and reformatted some functions.
moved tests from each files to tests/test_backtest file.
no new logic was added. Please take a look and review :)


algorithms/backtrack/generate_parenthesis.py
```diff
@@ -13,20 +13,21 @@
...
-    if right > 0:
-        add_pair(res, s+")", left, right-1)
-    if left > 0:
-        add_pair(res, s+"(", left-1, right+1)
-
 
-if __name__=="__main__":
-    print(gen_parenthesis(3))
+if __name__ == "__main__":
+    print(generate_parenthesis(3))
```
goswami-rahul: I think we should remove this `if __name__ == "__main__"` for consistency with the reet of the files.

algorithms/backtrack/letter_combination.py
```diff
@@ -7,7 +7,7 @@
 """
 
 
-def letter_combinations(digits:"str")->"List[str]":
+def letter_combinations(digits: str)->list[str]:
```
goswami-rahul: I think typing hint `List[str]` was correct. (don't know if `list[str]` is also possible). But for that we should also import `from typing import List`.

I suggest we annotate the types of args in the docstrings (the old way), mainly because `typing` is only supported by Python 3.5+, and also for consistency. :)

README.md
```diff
@@ -70,7 +70,7 @@ If you want to uninstall algorithms, it is as simple as:
 - [backtrack](algorithms/backtrack)
     - [general_solution.md](algorithms/backtrack/)
     - [anagram](algorithms/backtrack/anagram.py)
-    - [array_sum_combinations](algorithms/backtrack/array_sum_combinations.py)
+    - [array_sum_combinationss](algorithms/backtrack/array_sum_combinationss.py)
```
goswami-rahul: you got an extra __s__ there :)

algorithms/backtrack/__init__.py
```diff
@@ -0,0 +1,7 @@
+from .anagram import *
+from .array_sum_combinations import *
+from .combination_sum import *
+from .find_words import *
+from .pattern_match import *
+from .add_operators import *
+from .permute import *
```
goswami-rahul: Why did you skip some files here?

algorithms/backtrack/letter_combination.py
keon: i agree. will remove

algorithms/backtrack/__init__.py
keon: unincluded ones were the ones that were not tested.
But I included them now. Ill add more tests later

README.md
keon: thanks!

algorithms/backtrack/generate_parenthesis.py
keon: gotcha

PR #308, opened 2018-06-05
Created bogo_sort.py in algorithms/sort

9967han:
Although bogo_sorting is not useful for sorting, it can be used in training through comparison with the algorithms that actually used. It is also used as an example of logic type programming.

### DIFF
--- a/algorithms/sort/bogo_sort.py
+++ b/algorithms/sort/bogo_sort.py
@@ -1,0 +1,22 @@
+import random
+
+def bogo_sort(arr):
+    '''Bogo Sort
+        Best Case Complexity: O(n)
+        Worst Case Complexity: O(∞)
+        Average Case Complexity: O(n-n!)
+    '''
+    while not inorder(arr):
+        print(arr)
+        random.shuffle(arr)
+    return arr
+
+def inorder(arr):
+    #check the array is inorder
+    i = 0
+    arr_len = len(arr)
+    while i+1< arr_len:
+        if arr[i] > arr[i+1]:
+            return False
+        i += 1
+    return True
--- a/algorithms/sort/bogo_sort.py
+++ b/algorithms/sort/bogo_sort.py
@@ -4,19 +4,19 @@ def bogo_sort(arr):
     """Bogo Sort
         Best Case Complexity: O(n)
         Worst Case Complexity: O(∞)
-        Average Case Complexity: O(n*n!)
-    """"
-    while not inorder(arr):
-        print(arr)
+        Average Case Complexity: O(n(n-1)!)
+    """
+    def is_sorted(arr):
+        #check the array is inorder
+        i = 0
+        arr_len = len(arr)
+        while i+1 < arr_len:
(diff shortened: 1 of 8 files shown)


algorithms/sort/bogo_sort.py
```diff
@@ -0,0 +1,22 @@
...
+        Best Case Complexity: O(n)
+        Worst Case Complexity: O(∞)
+        Average Case Complexity: O(n-n!)
+    '''
+    while not inorder(arr):
+        print(arr)
+        random.shuffle(arr)
+    return arr
+
+def inorder(arr):
```
goswami-rahul: please rename it `is_sorted()` and define it inside the `bogo_sort` as its inner function (it shouldn't accessible outside the function).

algorithms/sort/bogo_sort.py
```diff
@@ -0,0 +1,22 @@
+import random
+
+def bogo_sort(arr):
+    '''Bogo Sort
```
goswami-rahul: use triple _double_ quotes `"""..."""`.

algorithms/sort/bogo_sort.py
```diff
@@ -0,0 +1,22 @@
...
+        Best Case Complexity: O(n)
+        Worst Case Complexity: O(∞)
+        Average Case Complexity: O(n-n!)
+    '''
+    while not inorder(arr):
+        print(arr)
```
goswami-rahul: remove `print` statement.

algorithms/sort/bogo_sort.py
```diff
@@ -0,0 +1,22 @@
...
+        random.shuffle(arr)
+    return arr
+
+def inorder(arr):
+    #check the array is inorder
+    i = 0
+    arr_len = len(arr)
+    while i+1< arr_len:
```
goswami-rahul: spaces around ` < `.

algorithms/sort/bogo_sort.py
```diff
@@ -0,0 +1,22 @@
+import random
+
+def bogo_sort(arr):
+    '''Bogo Sort
+        Best Case Complexity: O(n)
+        Worst Case Complexity: O(∞)
+        Average Case Complexity: O(n-n!)
```
goswami-rahul: what do you mean by `O(n - n!)` ?? Shouldn't this be `O(n * n!)` ?

tests/test_sort.py
```diff
@@ -16,6 +17,10 @@
 
 
 class TestSuite(unittest.TestCase):
+    def test_bogo_sort(self):
+        self.assertEqual([1, 5, 23, 57, 65, 1232],
```
goswami-rahul: I think we should use a small test case here (to reduce testing time) since it runs `O(n * n!)`. We should add a note in the description too, to avoid giving array with length > 6 (or 7 ?). It could take a lot of time for large `n`.

9967han:
Very Very thank you for your comment!!  :)

PR #309, opened 2018-06-05
Created factorial.py in algorithms/maths

wldn0202:
I also modified a TestCombination comment

### DIFF
--- a/algorithms/maths/factorial.py
+++ b/algorithms/maths/factorial.py
@@ -1,0 +1,19 @@
+#
+# This function calculates n!
+# Factorial function not works in less than 0
+#
+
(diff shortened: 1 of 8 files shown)


algorithms/maths/factorial.py
```diff
@@ -0,0 +1,19 @@
...
+
+    result = 1
+    for i in range(2, n+1):
+        result *= i
+
+    return result
+
+
+def factorial_recur(n):
+    if n==0:
```
goswami-rahul: please put spaces around those ` == `.

wldn0202:
I put space around "=="

Issue #293, opened 2018-06-01
Add Documentation


keon:
This is an extension of issue #264
To prepare for packaging, we need to properly explain how api works in our documentation.

goswami-rahul:
@keon should we show the API usage in each module's (or each file's) docstring, or do we write a separate API documentation?
We also need to have complete docstrings in each file to explain what problem the algorithm is solving.

keon:
@goswami-rahul 
Usually we write api usage in docstring and automatically compile it into docs.
And you're right. we have to write docstrings in the right format in each algorithm.

keon:
Closing — all modules now have docstrings, type hints, and complexity notes. README has one-line descriptions for every algorithm.

PR #314, opened 2018-06-06
Add documentation template

keon:
Empty documentation template for issue #293. 
added a little logo because why not


danghai:
@keon It is really awesome!

PR #316, opened 2018-06-06
Added link to OS Helpers, Build Status, and Coverage Status in README files.

wldn0202:
There were some README files that do not have these information in top of the page (English and German already had them), so I added them in Korean, Japanese and Chinese files.


danghai:
@wldn0202 Thanks PR

PR #317, opened 2018-06-06
Create bitonic_sort.py in /algorithms/sort/ and update test_sort, __init__ and all README files

hsi1032:
Create bitonic_sort.py in /algorithms/sort/ and update test_sort, __init__ and all README files
If anything wrong, please comment me
Thank you,

### DIFF
--- a/algorithms/sort/bitonic_sort.py
+++ b/algorithms/sort/bitonic_sort.py
@@ -11,14 +11,14 @@ def bitonic_sort(*args):
     """
     
     #input can be (arr, direction(inc or dec)) or just (arr) with default order(increase)
-    n = len(arr)
+    
     arr = args[0]
     if len(args)==2:
         direction = args[1]
     else:
         direction = True
         
-        
+    n = len(arr)
     def compare(arr, direction):
         n = len(arr)//2
         for i in range(n):
--- a/algorithms/sort/bitonic_sort.py
+++ b/algorithms/sort/bitonic_sort.py
@@ -32,7 +32,7 @@ def bitonic_sort(arr, reverse=False):
     if n <= 1:
         return arr
     # checks if n is power of two
-    if (n and (not(n & (n - 1))) ):
+    if not (n and (not(n & (n - 1))) ):
         raise ValueError("the size of input should be power of two")
     
     left = bitonic_sort(arr[:n // 2], True)
--- a/algorithms/sort/bitonic_sort.py
+++ b/algorithms/sort/bitonic_sort.py
@@ -29,13 +29,12 @@ def bitonic_sort(arr, reverse=False):
     
     #end of function(compare and bitionic_merge) definition
     n = len(arr)
-    
+    if n <= 1:
+        return arr
     # checks if n is power of two
     if (n and (not(n & (n - 1))) ):
         raise ValueError("the size of input should be power of two")
     
-    if n <= 1:
-        return arr
     left = bitonic_sort(arr[:n // 2], True)
     right = bitonic_sort(arr[n // 2:], False)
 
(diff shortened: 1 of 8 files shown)


algorithms/sort/bitonic_sort.py
```diff
@@ -0,0 +1,41 @@
+def bitonic_sort(arr, direction):
```
goswami-rahul: I guess `direction` is the flag used to indicate the ascending or descending. I think this should be renamed to something like `reverse`, and work just like built-in `sort` method.

Also make the second parameter optional (by having a default value of sorting in increasing order). It should be callable by just `bitonic_sort([1, 3, 4, 2])`.

algorithms/sort/bitonic_sort.py
```diff
@@ -0,0 +1,41 @@
...
+    reference: https://en.wikipedia.org/wiki/Bitonic_sorter
+    """
+    n = len(arr)
+    
+    def compare(arr, direction):
+        n = len(arr)//2
+        for i in range(n):
+            if direction == (arr[i] > arr[i+n]):
+                arr[i], arr[i+n] = arr[i+n], arr[i]
+        print(arr)
```
goswami-rahul: please remove the `print` statement

algorithms/sort/bitonic_sort.py
hsi1032: Thank you for your comment,
I change my code to `direction` to `reverse` with default value True,

algorithms/sort/bitonic_sort.py
hsi1032: Sorry, that's my mistake I delete it

algorithms/sort/bitonic_sort.py
```diff
@@ -0,0 +1,40 @@
...
+    It can sort only array that sizes power of 2
+    It can sort array in both increasing order and decreasing order by giving argument true(increasing) and false(decreasing)
+    
+    Worst-case in parallel: O(log(n)^2)
+    Worst-case in non-parallel: O(nlog(n)^2)
+    
+    reference: https://en.wikipedia.org/wiki/Bitonic_sorter
+    """
+
+    n = len(arr)
```
keon: placing this under the function definitions and edge case checks would make it more readable.

algorithms/sort/bitonic_sort.py
```diff
@@ -0,0 +1,40 @@
+def bitonic_sort(arr, reverse=False):
+    """
```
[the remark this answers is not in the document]
hsi1032: Thank you for pointing out my code's weakness
Do you mean that if array size is not power of 2, then return -1 or else because it returns wrong answer?

hsi1032:
My code is running well in my environment, python 2.7.12 and python 3.5.2,
but some check shows not successful, i.e. coverage/coveralls
Can I know what parts that I make mistakes?
Thank you for all reviewer,

algorithms/sort/bitonic_sort.py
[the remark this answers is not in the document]
keon: returning -1 would make it implicit.
but we want explicit error.
like so:
```python
# checks if n is power of two
if (n and (not(n & (n - 1))) ):
    raise ValueError("the size of input should be power of two")
```
you can just copy and paste the code above.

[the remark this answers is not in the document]
keon: my bad, it has to be `not (n and (not(n & (n - 1))) )` not `(n and (not(n & (n - 1))) )`

keon:
@hsi1032 no worries about the coverage test. Thanks!

PR #319, opened 2018-06-06
Create top_1.py, trimmean.py on array.  Changing arrays to array

aig031:
test_array, REAME files also arranged.


aig031:
I'm sorry to say that, I forgot to make another branches to make another PR. I will use another branches from now on.

goswami-rahul:
@aig031 Thanks for the reorganisation. We will need to change the README links too, `arrays` -> `array`.
And please keep short title of the PR. You can always add long description inside the description section :)

aig031:
@goswami-rahul Oh, I see. I also already changing README links on my another branch. Now, I just check if there is some mistakes.

algorithms/array/trimmean.py
```diff
@@ -0,0 +1,19 @@
...
+"""
+def trimmean(arr, per):
+    ratio = per/200
+    # /100 for easy calculation by *, and /2 for easy adaption to best and worst parts.
+    cal_sum = 0
+    # sum value to be calculated to trimmean.
+    arr.sort()
+    count = 0
+    # count how many values added.
+    for i in range(len(arr)):
```
keon: I could improve this by not just looping through all the elements, but instead use binary search to find the indexes we want to keep.
also, `+=` creates list every time, so `append()` is more efficient.
https://stackoverflow.com/questions/725782/in-python-what-is-the-difference-between-append-and/725882

aig031:
Thanks for reviews, I changed code, and rename code. However, then Coverage check doesn't give me successful checks. I don't know reason why... plz tell me why this happens.

goswami-rahul:
@aig031 you can ignore the checks by Coveralls, they basic calculate how much percentage of the code have been tested. If the percent goes down ( which is obvious if someone adds any code or adds any lines which don't need tests), the coverall shows unsuccessful check.

aig031:
Oh, I see, thanks for comments. It makes feel better.

algorithms/array/top_1.py
```diff
@@ -0,0 +1,36 @@
...
+
+    for i in arr:
+        if i in values:
+            values[i] += 1
+        else:
+            values[i] = 1
+
+    a = list(values.values())
+
+    a.sort()
```
keon: sort makes it at least `O(n log n)`
I also don't think this is needed

aig031: I truly sorry about miscalculated complexity. I forgot to calculate sort in complexity. I thought we need sort() to find f_val, however now I found that dic supports that function! It'll be corrected  ASAP.

keon:
no need to be sorry :) thats what the review is for.
I tried to resolve the conflict, but it seems everything got tangled up.
Could you push a new PR?

aig031:
Of course, I'll push a new PR

PR #323, opened 2018-06-06
WIP: Fix some issues in Travis and add more python env 

danghai:
This PR does some tasks:

- [x] : Add requirements dependency in `requirements.txt` for algorithms.

- [x] : Update cache, and coverall ( algorithms and tests) in Travis

- [x] : Switch to use `tox` for running the unittest, coveralls. Add more python environment for testing. It does not only test for python 3.6. It could probably test python 3.4, 3.5, 3.6

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,14 +3,14 @@ language: python
 cache:
   directories:
     - $HOME/.cache/pip
-python:
-    - 3.6
-    #- nightly
-    #- pypy3
 matrix:
-    allow_failures:
-        - python: nightly
-        - python: pypy3
+    include:
+        - python: 3.4
+          env: TOX_ENV=py34
+        - python: 3.5
+          env: TOX_ENV=py35,coverage
+        - python: 3.6
+          env: TOX_ENV=py36
 install:
     - pip install -r requirements.txt
 before_script:
@@ -21,17 +21,9 @@ before_script:
 script:
     # Check python install package
     - pip3 install -e .
-    # Check unittest running
-    - python3 -m unittest discover tests
-    # Check pytest running
-    - python3 -m pytest tests
-    # Run nose with coverage support
-    - nosetests --with-coverage
+    - tox -e $TOX_ENV
     # Check python uninstall package
     - pip3 uninstall -y algorithms
 notifications:
     on_success: change
     on_failure: change  # `always` will be the setting once code changes slow down
-
-after_success:
-  - coveralls
--- a/.travis.yml
+++ b/.travis.yml
@@ -11,10 +11,6 @@ matrix:
         - python: pypy3
 install:
     - pip install -r requirements.txt
-    - pip install flake8  # pytest  # add another testing frameworks later
-    - pip install python-coveralls
-    - pip install coverage
-    - pip install nose
 before_script:
     # stop the build if there are Python syntax errors or undefined names
     - flake8 . --count --select=E901,E999,F821,F822,F823 --show-source --statistics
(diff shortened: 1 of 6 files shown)


danghai:
It makes sure that Algorithms can successfully run under all version in python3. That is reason why I want to create more environment to test. However, there are some minor issue in it. You can check the travis to see what the issues are. In particular, it pass all tests for `python3.5`. 
But we got fail 1 case in `python3.4`: 
` self.assertEqual(unique_array_sum_combinations(A, B, C, target), answer)`
and `test_PriorityQueue` file. 
And similarly for `python3.6`. 
What do you @keon @goswami-rahul think about this idea? Could you help me fix the bugs?

goswami-rahul:
Here's the code for that function
```python
def unique_array_sum_combinations(A, B, C, target):
    """
    1. Sort all the arrays - a,b,c. - This improves average time complexity.
    2. If c[i] < Sum, then look for Sum - c[i] in array a and b.
       When pair found, insert c[i], a[j] & b[k] into the result list.
       This can be done in O(n).
    3. Keep on doing the above procedure while going through complete c array.
    Complexity: O(n(m+p))
    """
    def check_sum(n, *nums):
        if sum(x for x in nums) == n:
            return (True, nums)
        else:
            return (False, nums)

    pro = itertools.product(A, B, C)
    func = partial(check_sum, target)
    sums = list(itertools.starmap(func, pro))

    res = set()
    for s in sums:
        if s[0] is True and s[1] not in res:
            res.add(s[1])

    return list(res)

```
It uses `set` internally, which doesn't have any particular order. We should test with sorted list or make set of the list while testing equality.

As for the PriorityQueue error, it uses [collections.deque.insert](https://docs.python.org/3/library/collections.html#collections.deque.insert) which is new to Python 3.5. 

I think we should change that code to Python python 3.4 compatible, or is it fine to support Python 3.5+ ?? @danghai @keon

danghai:
In my opinion, I think the title of `Algorithms` from README: `Minimal and clean example implementations of data structures and algorithms in Python 3`. It should work fine all version of Python 3. If someone is using Python 3.4, they cannot run test command.

goswami-rahul:
But the `collections.deque.insert` failed in Python 3.6 test too. Its weird!

goswami-rahul:
>  It should work fine all version of Python 3.
Yeah, we would need some refactor in the PriorityQueue.

danghai:
@goswami-rahul Yeah it is weird! 
```
Traceback (most recent call last):
  File "/home/travis/build/danghai/algorithms/tests/test_queues.py", line 99, in test_PriorityQueue
    queue.push(1)
  File "/home/travis/build/danghai/algorithms/algorithms/queues/priority_queue.py", line 32, in push
    self.priority_queue_list.insert(index, node)
AttributeError: 'collections.deque' object has no attribute 'insert'
```
It is fail in both Python 3.4 and 3.6. But It passes Python 3.5 :)

goswami-rahul:
I have Python3.6 and it (obviously) works in my PC!
```python
In [6]: q
Out[6]: deque([1, 2, 3, 4])

In [7]: q.insert(1, 6)

In [8]: q
Out[8]: deque([1, 6, 2, 3, 4])
```

danghai:
Does it work with command: `python3 -m unittest tests.test_queues` ?

danghai:
@goswami-rahul I put comment in `PriorityQueue`, and then creating this issue #330  
One of the possible solution is we need to implement the `insert` method by using function in `collection.dequeue`. 
Travis should work fine now.

PR #333, opened 2018-06-07
Added simulation features in insertion & selection sorts

lg970325:


keon:
could you attach how the output looks?

lg970325:
For example, if list = [3, 1, 2, 4, 5]
If you execute insertion_sort(list,True),
[3,1,2,4,5]
[1,3,2,4,5]
[1,2,3,4,5]

If list = [5, 1, 2, 3, 4]
If you execute selection_sort(list,True),
[5,1,2,3,4]
[1,5,2,3,4]
[1,2,5,3,4]
[1,2,3,5,4]
[1,2,3,4,5]

lg970325:
@goswami-rahul I think your idea is good! I will add it soon. :)

algorithms/sort/insertion_sort.py
```diff
@@ -1,15 +1,24 @@
...
         Complexity: O(n^2)
     """
+    if simulation:
+        print(arr)
+        
     for i in range(len(arr)):
         cursor = arr[i]
         pos = i
+        
+        temp = arr[:]
```
keon: I believe this makes the space complexity from O(1) to O(N).
I'm not sure about increasing the time complexity for visualization purpose.

Would there be a better way we could do this?

goswami-rahul: Yeah, definitely there's a better way.
We can have a flag `swapped` to tell if any the position of any element has swapped.
for each `i` initialize `swapped = False`. Then if it enters the while loop and swaps, set `swapped = True`.
change the condition to
```python
if simulation and not swapped:
    print(arr)
```

goswami-rahul: @lg970325 But I don't understand why are not printing the array when its not changed.
I think we should print status of array after each `iteration`, irrespective of whether its changed or not, because then it will show the accurate steps of what the algorithm is doing.

lg970325: @goswami-rahul What I thought when implementing simulation feature was to show how the array **changed** as sort algorithm processes. That's the reason why I didn't printed the array when its not changed. But printing the array after each iteration is more accurate as you said, so I'll change that part.

lg970325: I changed bubble, insertion and selection sort to print the array at every iteration. So in insertion sort for example, if the arr=[3,4,5,1,2], and you execute insertion_sort(arr,True), you will see:
iteration 0 : [3, 4, 5, 1, 2]
iteration 1 : [3, 4, 5, 1, 2]
iteration 2 : [3, 4, 5, 1, 2]
iteration 3 : [3, 4, 5, 1, 2]
iteration 4 : [1, 3, 4, 5, 2]
iteration 5 : [1, 2, 3, 4, 5]

algorithms/sort/bubble_sort.py
```diff
@@ -27,6 +28,7 @@ def swap(i, j):
                 swap(i - 1, i)
                 swapped = True
                 if simulation:
-                    print(arr)
+                    iteration = iteration + 1
+                    print("iteration",iteration,":",arr)
```
goswami-rahul: could you print the array as
```python
print("iteration", iteration, ":", *arr)
```
so that we see the numbers without the `[]`.

lg970325: Changed :) 
arr --> *arr

lg970325:
@goswami-rahul Updated simulation feature in bubble sort in this PR :)

goswami-rahul:
oh sorry, didn't notice that :)

Issue #315, opened 2018-06-06
Simulation code for sort algorithms


geon0325:
I think it is a good idea to create a simulation code for sorting algorithms. 
For example, in case of Bubble sort, it will simulate like:
2 3 4 5 1
2 3 4 1 5
2 3 1 4 5
2 1 3 4 5 
1 2 3 4 5

wldn0202:
I want to participate in making simulation codes for sorting algorithms!

lg970325:
I have not written any simulation codes yet, but I will write it one by one from now :)

danghai:
Okay let's try to do one first.

psychozonestartshere:
Unable to import quick_sort from algorithms. Its showing ImportError:
 from algorithms import quicksort
ImportError: cannot import name 'quicksort' from 'algorithms' (C:\Users\encre\AppData\Local\Programs\Python\Python38-32\lib\site-packages\algorithms\__init__.py)
Even on the Documentation page of the Algorithms module, the link reference to the List of Implementations is not accessible.

keon:
Closing — visual simulations are out of scope for this library. The sorting module has comprehensive implementations with docstrings and complexity notes.

PR #342, opened 2018-06-09
simulation code about quick_sort and heap_sort

hsi1032:
With issue #315, I make quick sort and heap sort simulations.
However, these things actually not iterative form, but recursive form.
Then, I make simulation as changing is occured
(print after "partition" in quick sort / print after changing parent with child node in heap sort).

Thank you,

  - [x] Fixes #[issue number] if related to any issue


hsi1032:
I don't know why coverage failed, 
I check these simulations runs quite well in my computer

danghai:
@hsi1032 do not worry about coverage. It is fail because the test case does not cover/ run all line codes that you wrote

algorithms/sort/heap_sort.py
```diff
@@ -1,17 +1,21 @@
-def max_heap_sort(arr):
+def max_heap_sort(arr, simulation = False):
```
goswami-rahul: please remove the space around ` = ` in arguments.
```python
def max_heap_sort(arr, simulation=False):
```
Same changes in all function definitions.

hsi1032: I changed it,
Thank you for your comment! :)

PR #343, opened 2018-06-09
Created BFS maze_search.py and test case

lg970325:
I created maze search problem using BFS.
For example if the input grid is:
101111
101010
101011
111011
The minimum steps from location (0,0) to (3,5) is 14, so the output is 14.
Another example:
100
011
011
You cannot go from (0,0) to (2,2), so the output is -1.

  - [X] added links to it in the README files ?
  - [X] added description (overview of algorithm, time and space compleixty, and possible edge case) in docstrings ?


algorithms/bfs/maze_search.py
```diff
@@ -0,0 +1,30 @@
...
+    n = len(grid)
+    m = len(grid[0])
+    q = [(0,0,0)]
+    visit = [[0]*m for _ in range(n)]
+    if grid[0][0] == 0:
+        return -1
+    visit[0][0] = 1
+    while q:
+        i, j, step = q.pop(0)
+        if i == n-1 and j == m-1:
```
keon: you are doing edge-case checking here (line 21) and line 26.
I guess we could consolidate it and keep only the bottom one?

algorithms/bfs/maze_search.py
```diff
@@ -0,0 +1,30 @@
...
+            return step
+        for dir in range(4):
```
keon: dir is a keyword.
Please change it to something else.

keon:
oh and could you add time/space complexity? 
O(N) / O(N)

PR #354, opened 2018-06-12
remove `tests/*` from MANIFEST.in

### LINKED ISSUE #350: installing algorithms also installs tests
while installing through `pip install algorithms`, the `tests/` module also gets installed. I don't know if this will be of any use to the end-users. I think we can exclude this from getting packaged.  
Even if we want to package it with the install, the current api is 
```python
from tests import test_arrays
```
It should be inside the algorithms folder, to be explicit that these tests belong to algorithms, more like
```python
from algorithms.tests import test_arrays
```
What do you think @keon @danghai ?

goswami-rahul:
    - [x] Fixes #350

### DIFF
--- a/MANIFEST.in
+++ b/MANIFEST.in
@@ -1,4 +1,3 @@
 include README.md
 include LICENSE
 include algorithms/*
-include tests/*


goswami-rahul:
@keon @danghai  Could you also explain to me how this repo gets updated at the PyPI with a new version? I don't really understand how that works. Thanks.

keon:
@goswami-rahul Do you have PyPI account? I could add you there.

goswami-rahul:
@keon
Yeah, here's my PyPI account. https://pypi.org/user/goswami-rahul/

Issue #357, opened 2018-06-17
combination.py large number n, r case too slow to find a combination


arsgsg1:
hi, when i test combination.py, I found specific case with input large n, r too slow to find combination.

combination(50, 10) <- 10272278170 // too slow

in my opinion, using memoization method, can problem solving (using dictionary)

Can i send PR you?

huseyindalbudak:
you can use factorial and combination function. 
https://github.com/huseyindalbudak/mathpy/blob/master/combination.py
the code is so faster for combination(50, 10) than keon algorithms

PR #358, opened 2018-06-17
improvement combination.py using memoization

arsgsg1:
#357 I try to improve combination.py

- [X] **if done some changes :**
  - [X] Fixes #[issue number] if related to any issue

- [ ] **other**

### DIFF
--- a/tests/test_maths.py
+++ b/tests/test_maths.py
@@ -216,8 +216,8 @@ class TestCombination(unittest.TestCase):
         self.assertEqual(10, combination(5, 2))
         self.assertEqual(252, combination(10, 5))
     def test_combination_memo(self):
-        self.assertEqual(10272278170, combination_memo(50, 10, {}))
-        self.assertEqual(847660528, combination_memo(40, 10, {}))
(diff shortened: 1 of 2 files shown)


arsgsg1:
I forgot check parameter 
combination_memo(50, 10) <- before
combination_memo(50, 10) <- after

huseyindalbudak:
you can use factorial and combination function.
https://github.com/huseyindalbudak/mathpy/blob/master/combination.py
the code is so faster for combination(50, 10) than keon algorithms

tests/test_maths.py
```diff
@@ -215,7 +215,9 @@ class TestCombination(unittest.TestCase):
     def test_combination(self):
         self.assertEqual(10, combination(5, 2))
         self.assertEqual(252, combination(10, 5))
-        
+    def test_combination_memo(self):
+        self.assertEqual(10272278170, combination_memo(50, 10, {}))
```
goswami-rahul: We don't want to change the API, and passing an empty dictionary doesn't look good. We need an outside wrapper to make its API same as `combination`

PR #365, opened 2018-06-28
Implement n sum

iattempt:

- [x] **If creating a new file :**
  - [x] included tests with it ?
  - [x] added description (overview of algorithm, time and space compleixty, and possible edge case) in docstrings ?

- [ ] **other**


algorithms/arrays/n_sum.py
```diff
@@ -0,0 +1,97 @@
...
+Example:
+    Given n = 4, nums = [1, 0, -1, 0, -2, 2], target = 0,
+    return [[-2, -1, 1, 2], [-2, 0, 0, 2], [-1, 0, 0, 1]]
+"""
+
+__sum_closure = None
+__compare_closure = None
+
+
+def n_sum(n, nums, target, **kv):
```
goswami-rahul: In this repo, ideally all files have only one function in the global scope (that can be imported). Your implementation is good, but please change it a little bit as follows:
- I think we can drop the optional `keywords arguments`, since almost everytime, the default functions are used. This will reduce some complexity of the code.
- Most of the "hidden" functions `__*` can be just written as is, inside the main function body, specially `__n_sum`.
- If some extra functions are still required, make them as inner functions of `n_sum`.
- Also mention the Time Complexity if the solution. It not easy to follow this code.
Thanks.

algorithms/arrays/n_sum.py
```diff
@@ -0,0 +1,129 @@
...
+            return -1
+        elif num > taget:
+            return 1
+        else:
+            return 0
+
+    def same_closure_default(a, b):
+        return a == b
+
+    def n_sum(n, nums, target):
```
goswami-rahul: Wow, I am surprised we can define inner function with the same name as outer. I didn't know it was possible in Python, but it works! Thanks.

PR #367, opened 2018-06-28
Fixes #256 change strip_url_params3 to stop flaky unit test failures

### LINKED ISSUE #256: strip_url_params.py : BUG in function strip_url_params3(...)
The file ```strip_url_params.py``` is buggy. Because one times the tests run successful and a other times it fails. I comment out the test for this file.  

***important:*** The bug is in the function ```strip_url_params3``` of the file above.  

@SaadBenn Can you look in it? The project is from you.

programatt:
Use OrderedDict to preserve query string key order. This should prevent `?a=1&b=2` returning          
 `?b=2&a=1` sometimes and failing the test even though it's semantically correct.

- [x ] **if done some changes :**
  - [x ] wrote short description in the PR explaining what the changes do ?
  - [ x] Fixes #[issue number] if related to any issue

### DIFF
--- a/algorithms/strings/strip_url_params.py
+++ b/algorithms/strings/strip_url_params.py
@@ -10,6 +10,8 @@ from collections import defaultdict
 import urllib
 import urllib.parse
 
+from collections import OrderedDict
+
 # Here is a very non-pythonic grotesque solution
 def strip_url_params1(url, params_to_strip=None):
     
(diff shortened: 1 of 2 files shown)


danghai:
@programatt Thanks PR. It is really good. 
I agree with you I am not sure why coverage report something relating to `top_sort.py`

PR #373, opened 2018-07-07
fix longest_non_repeat_v1 and add test cases, fix __contains in Interval and make some code concise

newbility77:

     longest_non_repeat_v1 implementation has a bug. this PR is fixing it, please check the added test case.

- [ ] **other**

### DIFF
--- a/algorithms/arrays/longest_non_repeat.py
+++ b/algorithms/arrays/longest_non_repeat.py
@@ -21,7 +21,7 @@ def longest_non_repeat_v1(string):
     max_len = 0
     for i in string:
         if i in temp:
-            temp = []
+            temp = temp[temp.index(i) + 1:]
         temp.append(i)
         max_len = max(max_len, len(temp))
     return max_len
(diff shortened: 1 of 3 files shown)


algorithms/arrays/longest_non_repeat.py
```diff
@@ -21,7 +21,7 @@ def longest_non_repeat_v1(string):
     max_len = 0
     for i in string:
         if i in temp:
-            temp = []
+            temp = temp[temp.index(i) + 1:]
```
goswami-rahul: @newbility77 Nice catch! We could further optimize this code. Now it requires 2 traversals of `temp` in line`23` and `24`.
We can do in 1 traversal:
```py
for i in string:
    for idx, c in enumerate(temp):
        if c == i:
            temp = temp[idx + 1:]
            break
    temp.append(i)
    ...
```

danghai:
We resolve it ! Reopen when you have another way.

PR #374, opened 2018-07-09
Update limit.py

ggran-REN:
there are some syntax warnings in this code.
there maybe another logical error: 
in this code, when parameter min_lim and parameter max_lim are both None, it will not hit any if judgment, result will return []. 
however it should return arr[::] actually.

- [ ] **other**

### DIFF
--- a/algorithms/arrays/limit.py
+++ b/algorithms/arrays/limit.py
@@ -12,19 +12,25 @@ ex) limit([1,2,3,4,5], None, 3) = [1,2,3]
 Complexity = O(n)
 """
 
-def limit(arr, min_lim = None, max_lim = None):
+
+def limit(arr, min_lim=None, max_lim=None):
     result = []
-    if min_lim == None:
+    if min_lim is None and max_lim:
         for i in arr:
-            if i<= max_lim:
+            if i <= max_lim:
                 result.append(i)
-    elif max_lim == None:
+    elif min_lim and max_lim is None:
         for i in arr:
             if i >= min_lim:
                 result.append(i)
-    else:
+    elif min_lim and max_lim:
         for i in arr:
-            if i >= min_lim and i <= max_lim:
-                result.append(i)
+            if i < min_lim:
+                continue
+            if i > max_lim:
+                continue
+            result.append(i)
+    else:
+        result = arr[::]
 
     return result


algorithms/arrays/limit.py
```diff
@@ -12,19 +12,25 @@
...
+def limit(arr, min_lim=None, max_lim=None):
     result = []
-    if min_lim == None:
+    if min_lim is None and max_lim:
         for i in arr:
-            if i<= max_lim:
+            if i <= max_lim:
                 result.append(i)
-    elif max_lim == None:
+    elif min_lim and max_lim is None:
```
goswami-rahul: @ggran-REN The syntax warnings maybe in these `if` conditions. What if `min_lim` is `0`, then the code fails.  
Consider the case where I give `limit(arr, 0, 0)`, I would expect `[]`, but will get same array instead.  
All this can simply be solved by:
```py
def limit(arr, min_lim=None, max_lim=None):
    if min_lim is None:
        min_lim = float("-infinity")
    if max_lim is None:
        max_lim = float("infinity")
    result = []
    for i in arr:
        if min_lim <= i <= max_lim:
            result.append(i)
    return result
```

ggran-REN: Thasks alot ：-）！
i overlooked 0 is always False in python~

goswami-rahul: @ggran-REN I see that you are a first-time contributor, its great :).  
BTW, you can fix the code in your branch [`ggran-REN:patch-1`](https://github.com/ggran-REN/algorithms/tree/ggran-REN-patch-1)  to reflect changes in this PR. Thanks.

algorithms/arrays/limit.py
```diff
@@ -12,19 +12,25 @@
...
                 result.append(i)
-    else:
+    elif min_lim and max_lim:
         for i in arr:
-            if i >= min_lim and i <= max_lim:
-                result.append(i)
+            if i < min_lim:
+                continue
+            if i > max_lim:
+                continue
```
ankit167: nit
you could probably merge these two ifs into a single if statement.

ggran-REN: watch you tongue fool

keon: im confused

QuanDo2000:
There is currently a shorter implementation of limit.py in arrays so we should close this PR.

PR #375, opened 2018-07-16
PR to add some problem.

danghai:

- [x] **If creating a new file :**
  - [x] included tests with it ?
  - [x] added description (overview of algorithm, time and space compleixty, and possible edge case) in docstrings ?

- [ ] **other**


algorithms/search/next_greatest_letter.py
```diff
@@ -0,0 +1,37 @@
...
+Reference: https://leetcode.com/problems/find-smallest-letter-greater-than-target/description/
+'''
+
+import bisect
+
+def next_greatest_letter(letters, target):
+    index = bisect.bisect(letters, target)
+    return letters[index % len(letters)]
+
+def next_greatest_letter_v2(letters, target):
```
goswami-rahul: It will good to mention the time complexities here, as this one is linear.  
And maybe we should add a third version with `log n` with manual binary search (without using `bisect`)??

danghai:
@goswami-rahul thanks review, I add `big O`, and add `binary_search` using while ( not recursion).

goswami-rahul:
@danghai `bisect` runs in `O(log n)`. It is just a wrapper of binary search. See https://github.com/python/cpython/blob/68833dcbd45d30a779907db56d8fe0bb9c88f49a/Lib/bisect.py#L22-L41 .  
I will merge this. Thanks.

PR #397, opened 2018-09-03
fix sort/radix_sort (#395)

InnoFang:

    Use `position` as a loop variable instead of `is_done` to fix the bug that isn't to be find easily.
  - [x] Fixes #[issue number] if related to any issue
   #395 

- [ ] **other**

### DIFF
--- a/algorithms/sort/radix_sort.py
+++ b/algorithms/sort/radix_sort.py
@@ -29,3 +29,4 @@ def radix_sort(arr, simulation=False):
 
         position *= 10
     return arr
+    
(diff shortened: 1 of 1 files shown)


algorithms/sort/radix_sort.py
```diff
@@ -28,7 +25,7 @@ def radix_sort(arr, simulation=False):
 
         if simulation:
             iteration = iteration + 1
-            print("iteration",iteration,":",*arr)
+            print("iteration", iteration, ":", *arr)
 
         position *= 10
-    return arr
+    return arr
```
keon: Nit: could you add a new line at the end?

InnoFang: OK

PR #402, opened 2018-09-05
Update is_sorted.py

### LINKED ISSUE #401: Issue in the code of algorithms/algorithms/stack/is_sorted.py
Some changes need to be made in the code. 
1)The first 2 return statements have to be changed to break statements. This ensures that the code after the first for loop is executed. Previously it would never execute the latter part. 
2)The second append statement should add elements to stack and not storage_stack. Try the test case [3,4,7,8,5,6]. The program returns True. It should return False.

```
def is_sorted(stack):
    storage_stack = []
    for i in range(len(stack)):
        if len(stack) == 0:
            break    # return
        first_val = stack.pop()
        if len(stack) == 0:
            break    # return
        second_val = stack.pop()
        if first_val < second_val:
            return False
        storage_stack.append(first_val)
        stack.append(second_val)    # storage_stack.append(second_val)

    # Backup stack
    for i in range(len(storage_stack)):
        stack.append(storage_stack.pop())

    return True
```

anoubhav:
Changes made-
1) The first 2 return statements have been changed to break statements. This ensures that the code after the first for loop is executed. Previously it would never execute the latter part.
2) The second append statement should add elements to stack and not storage_stack. Try the test case [3,4,7,8,5,6]. The program returns True. It should return False.

  - [X] Fixes #401  if related to any issue

- [ ] **other**

### DIFF
--- a/algorithms/stack/is_sorted.py
+++ b/algorithms/stack/is_sorted.py
@@ -13,15 +13,15 @@ def is_sorted(stack):
     storage_stack = []
     for i in range(len(stack)):
         if len(stack) == 0:
-            return True
+            break
         first_val = stack.pop()
         if len(stack) == 0:
-            return True
+            break
         second_val = stack.pop()
         if first_val < second_val:
             return False
         storage_stack.append(first_val)
-        storage_stack.append(second_val)
+        stack.append(second_val)
 
     # Backup stack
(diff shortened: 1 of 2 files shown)


danghai:
@anoubhav Thanks PR, could you add your test case in `test_stack.py` in tests too?

PR #417, opened 2018-10-02
Optimize bubbe sort

VARoDeK:
- as after each step of 'outer-loop', one number reaches to its right position. Thus inner loop does not need to run for 'n' times in every cycle of 'outer-loop'.

- [ x] **if done some changes :**
  - [ x] wrote short description in the PR explaining what the changes do ?
  - [ x] Fixes #[issue number] if related to any issue

- [ ] **other**

### DIFF
--- a/algorithms/sort/bubble_sort.py
+++ b/algorithms/sort/bubble_sort.py
@@ -20,10 +20,10 @@ def bubble_sort(arr, simulation=False):
     iteration = 0
     if simulation:
         print("iteration",iteration,":",*arr)
-    x=-1
+    x = -1
     while swapped:
         swapped = False
-        x = x+1
+        x = x + 1
         for i in range(1, n-x):
             if arr[i - 1] > arr[i]:
                 swap(i - 1, i)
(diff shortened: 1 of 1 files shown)


algorithms/sort/bubble_sort.py
```diff
@@ -20,10 +20,11 @@ def swap(i, j):
     iteration = 0
     if simulation:
         print("iteration",iteration,":",*arr)
-    
+    x=-1
```
goswami-rahul: please use spaces before and after `=` following PEP guide.

algorithms/sort/bubble_sort.py
```diff
@@ -20,10 +20,11 @@ def swap(i, j):
...
     while swapped:
         swapped = False
-        for i in range(1, n):
+        x = x+1
+        for i in range(1, n-x):
```
goswami-rahul: give spaces here too.

algorithms/sort/bubble_sort.py
VARoDeK: Okay!

algorithms/sort/bubble_sort.py
VARoDeK: Doing it!

goswami-rahul: Looks good :+1:

PR #419, opened 2018-10-02
added lcs

utk3995:

- [ ] **other**

### DIFF
--- a/algorithms/dp/longest_common_subsequence.py
+++ b/algorithms/dp/longest_common_subsequence.py
@@ -1,0 +1,22 @@
+def lcs(X , Y): 
+    
+    l1 = len(X) 
+    l2 = len(Y) 
+  
+    dp = [[0]*(l2+1) for i in xrange(l1+1)] 
+  
+    for i in range(l1+1): 
+        for j in range(l2+1): 
+            if i == 0 or j == 0 : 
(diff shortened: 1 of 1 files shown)


algorithms/dp/longest_common_subsequence.py
```diff
@@ -0,0 +1,22 @@
...
+                dp[i][j] = 0
+            elif X[i-1] == Y[j-1]: 
+                dp[i][j] = dp[i-1][j-1]+1
+            else: 
+                dp[i][j] = max(dp[i-1][j] , dp[i][j-1]) 
+  
+    return dp[l1][l2]  
+
+
+X = "YRHHYDKP"
```
keon: can you add a unittest ?
Please take a look at [math unit tests](https://github.com/keon/algorithms/blob/master/tests/test_maths.py).
You can create a new file called `test_dp.py` for this as we have not created one yet.

algorithms/dp/longest_common_subsequence.py
```diff
@@ -0,0 +1,22 @@
+def lcs(X , Y): 
+    
+    l1 = len(X)
```
keon: I think we can use a better naming here.
also, could you provide some comment explaining the algo in a high level and what are the types of the arguments?

PR #422, opened 2018-10-05
add_stringOps

psusmit:
I added a simple palindrome algorithm for strings under stringOps folder under algorithms.
the algo is in python language


algorithms/stringOps/palindrome.py
```diff
@@ -0,0 +1,8 @@
+#@author susmit
+#program to check palindrone in python for strings and integer numbers
+
+def palindrome():
```
keon: please follow consistent naming method (filename)

Issue #337, opened 2018-06-07
copy_random_pointer.py needs rewritten


herinckc:
Path to file: /algorithms/linkedlist

At the moment, the copy_random_pointer.py file is entirely commented out. The comments indicate there is an intent to rewrite the file at some point; this issue report is simply intended to bring attention to this particular file in case it has been forgotten. Feel free to remove this issue if this is a low-priority task and/or someone is already working on it.

danghai:
@herinckc It is great when you can help us rewrite `copy_random_pointer.py` . Thanks

keon:
Closing — this has been addressed in the recent restructuring (GitHub Actions CI, ruff linting, updated README, PyPI v1.0.0, docstrings & type hints throughout).

PR #432, opened 2018-10-14
Fix array plus one v3

### LINKED ISSUE #431: Test fail. arrays/plus_one makes tests fail
The function plus_one_v3() tries to access the index as if it was an actual index (number) but since it's being used in a for loop with an enumerate, it is actually a tuple of (idx,value) for all the items in the list

I discovered this while making a PR for another issue since it would reject my build.

Already making a PR changing it so the tests run again.

krimvp:

- [X] **if done some changes :**
  - [X] Fixes #431 ad #337 

- [ ] **other**

### DIFF
--- a/algorithms/linkedlist/__init__.py
+++ b/algorithms/linkedlist/__init__.py
@@ -6,3 +6,4 @@ from .rotate_list import *
 from .is_cyclic import *
 from .merge_two_list import *
 from .is_palindrome import *
+from .copy_random_pointer import *
--- a/algorithms/linkedlist/copy_random_pointer.py
+++ b/algorithms/linkedlist/copy_random_pointer.py
@@ -5,7 +5,6 @@ pointer which could point to any node in the list or null.
 Return a deep copy of the list.
 """
 from collections import defaultdict
-from unittest import TestCase
 
 
 class RandomListNode(object):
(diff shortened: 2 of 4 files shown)


algorithms/linkedlist/copy_random_pointer.py
```diff
@@ -4,35 +4,90 @@
...
+
+
+class RandomListNode(object):
+    def __init__(self, label):
+        self.label = label
+        self.next = None
+        self.random = None
+
+
+class Solution0:
```
keon: you can unwrap this class and just leave it as a function

algorithms/linkedlist/copy_random_pointer.py
```diff
@@ -4,35 +4,90 @@
...
-  #   copy = defaultdict(lambda: RandomListNode(0))
-  #   copy[None] = None
-  #   node = head
-  #   while node:
-  #       copy[node].label = node.label
-  #       copy[node].next = copy[node.next]
-  #       copy[node].random = copy[node.random]
-  #       node = node.next
-  #   return copy[head]
+class Solution1:  # O(n)
```
keon: ditto

algorithms/linkedlist/copy_random_pointer.py
```diff
@@ -4,35 +4,90 @@
...
+        node = head
+        while node:
+            copy[node].label = node.label
+            copy[node].next = copy[node.next]
+            copy[node].random = copy[node.random]
+            node = node.next
+        return copy[head]
+
+
+class TestSuite(TestCase):
```
keon: could you move this to [tests folder](https://github.com/keon/algorithms/blob/master/tests/test_linkedlist.py)?

algorithms/linkedlist/__init__.py
```diff
@@ -6,3 +6,4 @@
 from .is_cyclic import *
 from .merge_two_list import *
 from .is_palindrome import *
+from .copy_random_pointer import *
```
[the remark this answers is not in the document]
krimvp: Oops my fault. Should be fine now

PR #487, opened 2019-03-09
 fix bugs, add descriptions, and add tests

JKLiang9714:
1. Fixed some bugs in "permute.py", "subsets.py"
2. Updated description for the algorithm in "palindrome_partitioning.py"
3. Added tests for the algorithm in "test_backtrack.py"


algorithms/backtrack/permute.py
```diff
@@ -46,7 +46,7 @@ def dfs(res, nums, path):
         if not nums:
             res.append(path)
         for i in range(len(nums)):
-            print(nums[:i]+nums[i+1:])
+            # print(nums[:i]+nums[i+1:])
```
keon: could you delete this line?

JKLiang9714: Got it. I have deleted it.

PR #498, opened 2019-04-22
Update limit.py

Everfighting:
- [X] **Consider boundary conditions**

### DIFF
--- a/algorithms/arrays/limit.py
+++ b/algorithms/arrays/limit.py
@@ -13,13 +13,13 @@ Complexity = O(n)
 """
 
 # tl:dr -- array slicing by value
-def limit(arr, min_lim = None, max_lim = None):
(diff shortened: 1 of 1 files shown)


Everfighting:
> Thanks for the bug fix, minor comments added.
 
Sorry, this is my first pull-request, I don't know whether my operation is right?  
I reopen it ,and add the fix following your suggestions

travishen:
> > Thanks for the bug fix, minor comments added.
> 
> Sorry, this is my first pull-request, I don't know whether my operation is right?
> I reopen it ,and add the fix following your suggestions

Hi @Everfighting, looks like you have two commit now, my suggestion is that you could use `git rebase` to squash these two commit and force update to this branch.

keon:
its fine, i can squash them while merging into master.

PR #531, opened 2019-10-01
Added Segment Trees

adi0311:
1. [SegmentTree.py] Segment Tree implementation for Sum of a particular range and update an element in an array with proper comments. 
2. [SegmentTree2.py] Segment Tree implementation for Minimum in a particular range and update an element in an array with proper comments.

### DIFF
--- a/algorithms/tree/segment_tree/SegmentTree.py
+++ b/algorithms/tree/segment_tree/SegmentTree.py
@@ -1,0 +1,66 @@
+class SegmentTree:
+    def __init__(self, values):
+        self.valarr = values
+        self.arr = dict()
+
+    #start is the starting index of node.
+    #end is the ending index of node.
+    #l is the lower bound of given query.
+    #r is the upper bound of given query.
+
+    def buildTree(self, start, end, node):
+        if start == end:
+            self.arr[node] = self.valarr[start]
+            return
+        mid = (start+end)//2
+        #Building the left subtree of the node.
+        self.buildTree(start, mid, node*2)
+        #Building the right subtree of the node.
+        self.buildTree(mid+1, end, node*2+1)
+        #Assign the value of node as the sum of its children.
+        self.arr[node] = self.arr[node*2]+self.arr[node*2+1]
+
+    def rangeQuery(self, node, start, end, l, r):
+        #When start and end index of the given node lies between the query range[l, r].
+        if (l <= start and r >= end):
+            return self.arr[node]
+        #When the start and end index of the given node lies completely outside of the query range[l, r].
+        if (end < l or start > r):
+            return 0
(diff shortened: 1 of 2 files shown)


algorithms/tree/segment_tree/SegmentTree.py
```diff
@@ -0,0 +1,66 @@
...
+    def __init__(self, values):
+        self.valarr = values
+        self.arr = dict()
+
+    #start is the starting index of node.
+    #end is the ending index of node.
+    #l is the lower bound of given query.
+    #r is the upper bound of given query.
+
+    def buildTree(self, start, end, node):
```
keon: could you make the function, variable and file names snake case (ex `build_tree`)?

algorithms/tree/segment_tree/SegmentTree.py
```diff
@@ -0,0 +1,66 @@
...
+        if start <= position <= end:
+            self.arr[node] += (newvalue-oldvalue)
+        #Updating all those nodes where position lies within its start and end index.
+        if start != end:
+            mid = (start+end)//2
+            self.update(node*2, newvalue, oldvalue, position, start, mid)
+            self.update(node*2+1, newvalue, oldvalue, position, mid+1, end)
+
+#Code to run the above functions
+if __name__ == '__main__':
```
keon: it should be moved to unit tests

algorithms/tree/segment_tree/SegmentTree2.py
```diff
@@ -0,0 +1,56 @@
...
+    def update(self, node, newvalue, position, start, end):
+        if start <= position <= end:
+            self.minarr[node] = min(self.minarr[node], newvalue)
+        # Updating all those nodes where position lies within its start and end index.
+        if start != end:
+            mid = (start + end) // 2
+            self.update(node * 2, newvalue, position, start, mid)
+            self.update(node * 2 + 1, newvalue, position, mid + 1, end)
+
+arr = [10, 5, 9, 3, 4, 8, 6, 7, 2, 1]
```
keon: these should be moved to unit tests

adi0311: Moved the testing conditions to unit tests.

algorithms/tree/segment_tree/SegmentTree.py
adi0311: Functions, variables, classes and files name changed to snake case instead of camel case.

algorithms/tree/segment_tree/SegmentTree.py
adi0311: Moved the above statement to unit test.

keon:

PR #611, opened 2019-12-07
Add kmp

sonho00:

  - [X] added links to it in the README files ?
  - [X] added description (overview of algorithm, time and space complexity, and possible edge case) in docstrings ?

- [ ] **other**

### DIFF
--- a/tests/test_strings.py
+++ b/tests/test_strings.py
@@ -541,6 +541,7 @@ class TestKnuthMorrisPratt(unittest.TestCase):
         unittest {[type]} -- [description]
     """
     def test_knuth_morris_pratt(self):
+        self.assertEqual([0, 1, 2, 3, 4], knuth_morris_pratt("aaaaaaa", "aaa"))
         self.assertEqual([0, 4], knuth_morris_pratt("abcdabc", "abc"))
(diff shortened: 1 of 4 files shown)


tests/test_strings.py
```diff
@@ -532,5 +533,15 @@ def test_atbash_cipher(self):
...
+class TestKnuthMorrisPratt(unittest.TestCase):
+    """[summary]
+    Test for the file knuth_morris_pratt.py
+
+    Arguments:
+        unittest {[type]} -- [description]
+    """
+    def test_knuth_morris_pratt(self):
+        self.assertEqual([0, 4], knuth_morris_pratt("abcdabc", "abc"))
+
```
keon: think more about possible edge cases. what if the string does not match? what if more than 1 possible matches?

sonho00: Sorry, the description of the algorithm was quite unkind. And I edited it now. If the string does not match, this code returns an empty list. If there are more than 1 cases, list of the start index will be returned.

keon: Sorry if I wasn't clear. I meant unit tests should cover those edge cases.

sonho00: I added test case for those cases. Can you check it?

keon: thanks!

PR #615, opened 2019-12-09
Add diffie hellman key exchange

D-Sinus:

  - [X] added description (overview of algorithm, time and space complexity, and possible edge case) in docstrings ?

- [ ] **other**


D-Sinus:
To implement DH key exchange, prime_check.py (written by 'goswami-rahul' and 'Hai Honag Dang') and euler_totient function (written by 'goswami-rahul') is needed.
So I copy their code and write comment about that (this is not my code).
And also this requires algorithms that I commit before (find order/primitive root).
That's why the length of code of this algorithm is quite long.

Again, Please let me know if such usage is problematic.
Thanks.

algorithms/maths/diffie_hellman_key_exchange.py
```diff
@@ -0,0 +1,183 @@
...
+            p_root_list.index(a)
+        except ValueError as e:
+            print("%d is not a primitive root of %d" % (a, p))
+            return -1
+            """ a must be primitive root of p """
+        
+        a_pr_k = alice_private_key(p)
+        a_pu_k = alice_public_key(a_pr_k, a, p)
+        print ("Private key of Alice = %d" % a_pr_k)
+        print ("Public key of Alice = %d" % a_pu_k)
```
keon: could you make the print statements optional?

algorithms/maths/diffie_hellman_key_exchange.py
```diff
@@ -0,0 +1,183 @@
...
+        print("%d is not a prime number" % p)
+        return -1
+        """p must be large prime number"""
+    else:
+        try:
+            p_root_list = find_primitive_root(p)
+            p_root_list.index(a)
+        except ValueError as e:
+            print("%d is not a primitive root of %d" % (a, p))
+            return -1
```
keon: please raise error in this case so we can only return bool type.
If a function returns two types (-1, bool) the caller needs to branch out by the return type.

D-Sinus: [Edit] For primitive root case, code now return False, for consistency

algorithms/maths/diffie_hellman_key_exchange.py
D-Sinus: [Edit] Add optional parameter, and those statements only appears when that parameter is given.

I left shared-key print statements for comparison.
Do you think those statements should be optional also?
Please let me hear your opinion. Thanks.

keon:
thanks so much for the contribution!

PR #622, opened 2019-12-17
Update flatten.py

sunshine0123:
use a  Recursion function


algorithms/arrays/flatten.py
```diff
@@ -29,3 +29,16 @@ def flatten_iter(iterable):
             yield from flatten_iter(element)    
         else:
             yield element
+
+    
+############## use a Recursion function ###########
+def flattenArray(A):
```
keon: please use snake case

keon:

PR #636, opened 2020-02-19
Add two more breaking bad tests

JohanKJIP:
Add two more tests to `bracket(words, symbols)` function in `strings/breaking_bad.py`. Adding these tests results in 100% branch coverage for the function. 

- Test 1: Test to match symbols that do not exist in the word list. This should return the words without any changes made to them.

- Test 2: Try to match two duplicate symbols. It should only encapsulate the match once ([i]), not twice ([[i]]).


ericklarac:
Tests are okay, but I think the cases should be added on the ```python test_bracket``` function.

keon:
@ericklarac feel free to merge yourself :)

ericklarac:
@keon Thanks, actually I thought you have to review it.

PR #637, opened 2020-02-20
Add tests to test_linkedlist to increase coverage for is_palindrome_dict

emmariaolsson:
Add three more assertions with corresponding lists in setUp() to test the function is_palindrome_dict in linkedlist/is_palindrome.py. Adding these test cases increases the branch coverage for the function from 77% to 100%.

**List of added test cases:**
1. A linked list consisting of an odd number of nodes and is _not_ a valid palindrome. This assertion expects the function to return False.

2. A linked list consisting of _only one_ node. This assertion expects the function to return True.

3. A case when there is no "head" at all and the input is _None_. This assertion expects the function to return True.

### DIFF
--- a/tests/test_linkedlist.py
+++ b/tests/test_linkedlist.py
@@ -9,7 +9,8 @@ from algorithms.linkedlist import (
     is_cyclic,
     merge_two_list, merge_two_list_recur,
     is_palindrome, is_palindrome_stack, is_palindrome_dict,
-    RandomListNode, copy_random_pointer_v1, copy_random_pointer_v2
+    RandomListNode, copy_random_pointer_v1, copy_random_pointer_v2,
+    intersection
 )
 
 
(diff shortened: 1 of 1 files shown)


tests/test_linkedlist.py
```diff
@@ -210,6 +224,72 @@ def _init_random_list_nodes(self):
...
+        a1.next = b1
+        b1.next = c1
+        a2.next = b2
+        b2.next = c2
+
+        self.assertEqual(None, intersection.intersection(a1, a2))
+
+    def test_intersection_one_longer(self):
+        # create linked list as:
+        # 1 -> 3 -> 5
```
ericklarac: Add the node correspondig to 0.

```
 # 0 -> 1 -> ...
```

keon:

PR #680, opened 2020-07-21
Creating Program to remove duplicates

Ratna04priya:
Adding Program to remove duplicates in Python

### DIFF
--- a/algorithms/strings/remove_duplicate.py
+++ b/algorithms/strings/remove_duplicate.py
@@ -1,0 +1,11 @@
+def remove_dups(string):
+    res = string[0]
+    for ch in string:
+        if res[-1] != ch:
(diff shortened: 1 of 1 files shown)


algorithms/strings/remove_duplicate.py
```diff
@@ -0,0 +1,11 @@
+def remove_dups(string):
```
ankit167, line 1: the method name looks a bit misleading. If I understand correctly, the program removes adjacent duplicates (rather than removing duplicates)
Input: `faacaacccgfs`
Output: `facacgfs`

Ratna04priya: yeah, right.. you got that.

keon:

PR #693, opened 2020-08-24
updated flatten.py

Uddeshya-1812:
Updated flatten.py ..removed isinstance method and now char arrays can also be reduced to 1-D array.

### DIFF
--- a/algorithms/arrays/flatten.py
+++ b/algorithms/arrays/flatten.py
@@ -3,29 +3,27 @@ Implement Flatten Arrays.
 Given an array that may contain nested arrays,
 produce a single resultant array.
 """
-from collections.abc import Iterable
-
 
 # return list
 def flatten(input_arr, output_arr=None):
-    if output_arr is None:
-        output_arr = []
-    for ele in input_arr:
-        if not isinstance(ele, str) and isinstance(ele, Iterable):
-            flatten(ele, output_arr)    #tail-recursion
-        else:
-            output_arr.append(ele)      #produce the result
-    return output_arr
+	if output_arr is None:
+		output_arr = []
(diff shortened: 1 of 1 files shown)


algorithms/arrays/flatten.py
```diff
@@ -3,29 +3,27 @@
...
-        else:
-            output_arr.append(ele)      #produce the result
-    return output_arr
+	if output_arr is None:
+		output_arr = []
+	for ele in input_arr:
+		try:
+			if len(ele) > 0:
+				flatten(ele, output_arr)    #tail-recursion
+		except:output_arr.append(ele)      #produce the result
```
goswami-rahul, line 15: Don't prefer bare excepts.

algorithms/arrays/flatten.py
```diff
@@ -3,29 +3,27 @@
...
-    if output_arr is None:
-        output_arr = []
-    for ele in input_arr:
-        if not isinstance(ele, str) and isinstance(ele, Iterable):
-            flatten(ele, output_arr)    #tail-recursion
-        else:
-            output_arr.append(ele)      #produce the result
-    return output_arr
+	if output_arr is None:
+		output_arr = []
```
goswami-rahul, line 10: Don't use tabs. Prefer 4 spaces.

algorithms/arrays/flatten.py
```diff
@@ -3,29 +3,27 @@
...
+	for ele in input_arr:
+		try:
+			if len(ele) > 0:
```
goswami-rahul, line 13: Not every object that has `__len__` implemented is `iterable`. I don't see how this improves the existing method.

PR #694, opened 2020-08-29
Add 2-3-4 Tree Structure Algorithms

TseChapman:
I implement the 2-3-4 Tree structure in Python. This is my first time contributing in open source, so be welcome to give me professional comment. Thank you.

This Tree structure was taught to me in C++ and now I have implemented it in Python.
The Folder included a input .txt file, where I randomly write out a tons of number, some are duplicated.

When you run the .py file, you will get result of many unsuccessful insert. This is because of duplicates key insert.

At last, you will also get 2 printed lines. One is the root node's A value. Second the result of search a random choice of number to search.
By searching the root node's A value, you should see "Tree234: search(key): Successfully find the key by searching 1 node"


goswami-rahul (changes requested):
Hi, thanks for contributing. We don't need that tree input.txt file. You can add a small tree for testing in `tests/test_tree.py`. Also remove the print statements from the algorithm, you can replace those with just comments.

TseChapman:
@goswami-rahul Thanks for the comment, I will change it and push it again. Thanks

PR #704, opened 2020-10-01
Added binomial coefficient calculator in algorithms/maths

DarkMoron:


ericklarac:
Could you add test cases?

DarkMoron:
> Could you add test cases?

They are already there at the end of the .cpp file.

goswami-rahul:
A different version is already done at [`maths/combination.py`](https://github.com/keon/algorithms/blob/master/algorithms/maths/combination.py). You can it as another function there. Also, don't perform any interaction (input, print) here, add the tests at `tests/`.

keon:

PR #711, opened 2020-10-02
Added Matrix Exponentiation

c-utkarsh:


algorithms/matrix/__init__.py
```diff
@@ -1 +1 @@
-
+from .matrix_exponentiation import *
```
goswami-rahul, line 1: We don't need this. `multiply` collides with `multiply.py`.

algorithms/matrix/matrix_exponentiation.py
```diff
@@ -0,0 +1,60 @@
...
+    elif n % 2 == 1:
+        return multiply(matrix_exponentiation(mat, n - 1), mat)
+    else:
+        tmp = matrix_exponentiation(mat, n // 2)
+        return multiply(tmp, tmp)
+
+if __name__ == "__main__":
+    mat = [[1, 0, 2], [2, 1, 0], [0, 2, 1]]
+    
+    res0 = matrix_exponentiation(mat, 0)
```
goswami-rahul, line 46: Add tests at [`tests/test_matrix.py`](https://github.com/keon/algorithms/blob/master/tests/test_matrix.py) similar to other tests.

goswami-rahul (changes requested):

c-utkarsh:
Hi @goswami-rahul, I've incorporated the requested changes.

goswami-rahul (approved):
Thanks. LGTM

PR #735, opened 2020-10-31
Full binary tree from preorder and postorder

ShimoniSinha2019H1030019G:
I have created an implementation that constructs a full binary tree from given preorder and postorder traversals of the tree.
The file has been stored in the tree folder.

PR #736, opened 2020-11-01
Construct full binary tree from preorder and postorder traversal

ShimoniSinha2019H1030019G:
I have created an implementation that constructs a full binary tree from given preorder and postorder traversals of the tree.
The file has been stored in the tree folder. 

It is modification of PR #735 with unit test also added.


ericklarac:
@ShimoniSinha2019H1030019G Use snake case to name the functions and the variables. Also, move the unit test to `tests/test_tree.py`

ShimoniSinha2019H1030019G:
@ericklarac I have moved the unit test to tests/test_tree.py and have also modified the variable names and function names as per snake case.

ShimoniSinha2019H1030019G:
@ericklarac  Thankyou very much.

PR #659, opened 2020-06-01
Create num_digits.py

vicky1999:
num_digits.py was created to find the  number of digits in a number using log10 method in math module.

PR #740, opened 2020-11-17
finding number of digits using logs created with unit tests

vicky1999:
num_digits.py was created to find the number of digits in a number using log10 method in math module.
This PR was created from #659 

Implementation and unit tests were done.


algorithms/maths/num_digits.py
```diff
@@ -0,0 +1,10 @@
+"""
+   num_digits() method will return the number of digits of a number in O(1) time using math.log10() method.
+"""
+
+import math
+
+def num_digits(n):
+  if(n==0):
```
ankit167, line 8: Please handle the case if n < 0

tests/test_maths.py
```diff
@@ -448,5 +450,18 @@ def test_find_order_simple(self):
...
+    """[summary]
+    Test for the file num_digits.py
+
+    Arguments:
+        unittest {[type]} -- [description]
+    """
+    def test_num_digits(self):
+        self.assertEqual(2,num_digits(12))
+        self.assertEqual(5,num_digits(99999))
+        self.assertEqual(1,num_digits(8))
```
ankit167, line 463: Add test case for:
1. n = 0
2. n < 0

ankit167 (changes requested):

tests/test_maths.py
vicky1999: Sure.  I'll do that

vicky1999:
the test case for n<0 has been handled and the unit tests were created for:
- n=0
- n<0

ankit167 (approved):

PR #741, opened 2020-11-21
Add Polynomials (DS) + Unit tests, and Cycle Index Polynomial for Symmetry Groups (Algorithm)

aalekhpatel07:
# New Data Structures
## Monomial and Polynomial
Supports operations like:
- Add
- Subtract
- Multiply
- Divide*
- Invert*
- Evaluate at given values

### It is recommended to read the docstrings for a clear idea of how to use these structures.
---

# New Algorithm
## Cycle Index Polynomial for Symmetry Groups
- Generate a polynomial for Symmetry Group of arbitrary order.
- Various advanced counting problems have their solutions spring out of the Cycle Index of some Group.
- Here's a [very good resource](http://www-groups.mcs.st-andrews.ac.uk/~pjc/Teaching/MT5821/2/w8.pdf) to understand the Combinatorics behind it: 

**Now that there is support for polynomials (technically Multinomials), one can be motivated to add other algorithms for producing well-known multinomials. Hence, there is plenty of room for advancement.**

- For example: One can add algorithms to compute cycle index of other groups like the Cyclic Groups, Alternating Group, etc.


ericklarac:
Nice PR, could you add your implementation on the readme?

aalekhpatel07:
> Nice PR, could you add your implementation on the readme?

Done! :) Requesting to merge.

PR #751, opened 2021-01-15
Added implementation of the  algorithm without using recursion.

LuckyDenis:
Good afternoon!
I would like to suggest that you add an implementation of the "smoothing" algorithm without using recursion.

Added:
[x] Flatter algorithm used stack
[x] Tests


keon:

PR #752, opened 2021-01-17
Create longest_sum_contigous_array.py

DivyanshMandhan:


ericklarac:
This is Kadane's Algorithm. Given your same example [-2,-3,4,-1,-2,1,5,3], the longes sum contiguous array is [4,-1,-2,1,5,3] which sums 10. Also this algorithm is [already implemented](https://github.com/keon/algorithms/blob/master/algorithms/dp/max_subarray.py).

PR #778, opened 2021-05-12
fix_rotate.py

TrentonSouth:
The way it was, you couldn't pass in a rotation higher than the length of the string doubled without getting a blank string in return. This way allows you to pass in any positive integer and get a result.

Also,
The last line of the description said that it took 2 strings and returned a boolean. This is not correct. The first line of the comment is correct.


ericklarac:
Nice fix, could you update the algorithm description, please?

PR #787, opened 2021-06-12
Update __init__.py

123456789x:
There are about 10 or so new algorithms that were not included in the ```__init__.py``` file, which caused a no attribute error in Python.


keon:

PR #801, opened 2021-08-30
Added graph coloring algorithm inside Greedy 

ritikverma2000:
**Graph coloring problem is to assign colors to certain elements of a graph subject to certain constraints. Vertex coloring is the most common graph coloring problem. The problem is, given m colors, find a way of coloring the vertices of a graph such that no two adjacent vertices are colored using the same color.
Chromatic Number: The smallest number of colors needed to color a graph G is called its chromatic number.
So basically my code finds the chromatic number for the given graph.**

### DIFF
--- a/README.md
+++ b/README.md
@@ -51,7 +51,8 @@ If you want to uninstall algorithms, it is as simple as:
     $ pip3 uninstall -y algorithms
 
 ## List of Implementations
-
+- [Greedy](algorithms/arrays)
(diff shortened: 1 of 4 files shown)


ritikverma2000:
@ericklarac if you could pls check my code and merge it in your repo it would be great as it an urgent requirement for me.

keon:

PR #825, opened 2022-01-22
code cleanup for algorithms

shimafallah:
I made a series of changes in the array and search and stack. 
I hope you like it

### DIFF
--- a/algorithms/search/search_insert.py
+++ b/algorithms/search/search_insert.py
@@ -12,7 +12,7 @@ def search_insert(array, val):
     low = 0
     high = len(array) - 1
     while low <=  high:
-        mid = low + (high - low) // 2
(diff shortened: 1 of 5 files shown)


algorithms/search/search_insert.py
```diff
@@ -12,7 +12,7 @@ def search_insert(array, val):
     low = 0
     high = len(array) - 1
     while low <=  high:
-        mid = low + (high - low) // 2
+        mid = low + (high % 2)
```
ankit167, line 15: I don't see this as the right way to find mid index

e.g.: low = 0, high = 7
```
old way: 0 + (7-0)//2 = 3 // correct
new way: 0 + (7%2) = 1 // incorrect
```

PR #827, opened 2022-01-30
Add direction change level order

hugochapado:
[Change Direction LvlOrder.txt](https://github.com/keon/algorithms/files/7967095/Change.Direction.LvlOrder.txt)

This is a level Order Traversal of A tree But with the posibility of changing the Direction of it between The amount of levels Someone Wants. In the code the direction will change every 2 levels(form left to right and viceversa), But it can be changed by modifying the value of variable count


keon:
Thanks! The PR scope is too broad with many unrelated changes. Closing.

PR #830, opened 2022-02-16
Add UnionFind data structure with a Rust backend.

aalekhpatel07:
Feature:

- A fast implementation of the UnionFind data structure that has a [Rust backend](https://docs.rs/reunion/0.1.14/reunion/).

Minor updates:

- Refactor the Solution class in `algorithms/unionfind/count_islands.py` to use the new UnionFind API.

Bug Fixes:

- Add None checks at the boundaries for Array limit (so that the tests pass).


keon:
Thanks! Adding external Rust dependencies is out of scope for this pure-Python library. Closing.

PR #831, opened 2022-02-16
Update pytest command.

aalekhpatel07:
Invoke pytest as a python module rather than a standalone executable.


ankit167 (approved):

aalekhpatel07:
Note that the build fails because there is a failing test:

> FAILED tests/test_array.py::TestLimit::test_limit - TypeError: '<=' not suppo...

#830 fixes this test and this actions workflow should start working again post merge.

PR #832, opened 2022-02-16
Bug fix: Add None checks for the boundary values.

aalekhpatel07:
Its bugging me that CI workflow is failing because of a test for this function. This bug fix should bring green ticks back again. ;)


aalekhpatel07:
@ankit167 This is a quick bug fix for the failing test that is breaking the workflow. Consider merging this one in and then we can spend some time to discuss about #830 on another day.

ankit167 (approved):

PR #840, opened 2022-03-03
Polonomial division

nolanderc:
This is a fix to #839. 

Division is implemented using euclidean division (aka. long division) and generalized to polynomials with multiple variables using a Gröbner-basis. The choice of basis was made by a loose recommendation in [this article](https://en.wikipedia.org/wiki/Monomial_order#Graded_reverse_lexicographic_order).


keon:

Issue #480, opened 2019-02-15
Add Bitmasking in Dp


adMenon:
Bitmasking and DP is a method used for solving questions like assign unique caps among persons, etc.
Can I add this algorithm?

keon:
Closing — this algorithm is now implemented in the repo. Thanks for the suggestion!

PR #855, opened 2022-03-07
Bitmasking algorithms using DP

AmandaStromdahl:
We implemented two bitmasking algorithms for issue #480, as part of a course at KTH. The first algorithm solves the problem of assigning unique caps to a set of people, and the second algorithm implements a solution to the TSP.


keon:

PR #862, opened 2022-04-19
Added more greedy techniques

PalashHawee:
My little effort contributed to two more problems solutions in Greedy algorithms.


keon:
Thanks for the greedy technique additions\! The implementations need tests and don't match the current code style. Closing.

PR #866, opened 2022-05-23
Structure testing function num_decodings.py

Eurus44:


aminulqrshah:
def binary_tree_paths(root): 
     res = [] 
     if root is None: 
         return res 
     dfs(res, root, str(root.val)) 
     return res 
  
  
 def dfs(res, root, cur): 
     if root.left is None and root.right is None: 
         res.append(cur) 
     if root.left: 
         dfs(res, root.left, cur+'->'+str(root.left.val)) 
     if root.right: 
         dfs(res, root.right, cur+'->'+str(root.right.val))@

keon:
Thanks for the test additions! The test suite was overhauled in v1.0.0. Closing as stale.

PR #867, opened 2022-06-29
Add Exponential Search

yuviii99:
Added exponential search algorithm which is an improvised version of binary search algorithm implemented in python.


keon:

PR #871, opened 2022-07-10
Adding linear regression algorithm

MakanFar:
Implementing linear regression in python from scratch.


keon:

PR #880, opened 2022-08-19
Added a more efficient approach which take 2 args.

patilabhay679:


keon:
Thanks! The existing binary search implementation is preferred. Closing.

PR #881, opened 2022-08-29
Adding Non commutative version

sdt4936:
This is the non-commutative version for segment tree.


sdt4936:
@keon Lemme know if this looks good to merge!

keon:
Thanks! Adding external Rust dependencies is out of scope for this pure-Python library. Closing.

PR #886, opened 2022-12-16
modified the return from both main and helper.

patilabhay679:
It looks like there is no need to create a copy, we can replace the new elements inplace by mutating the contents of array.


algorithms/sort/merge_sort.py
```diff
@@ -35,4 +37,4 @@ def merge(left, right, merged):
         merged[left_cursor + right_cursor] = right[right_cursor]
 
     # Return result
-    return merged
+    # return merged # do not return anything, as it is replacing inplace.
```
ericklarac, line 40: Remove comments instead

PR #889, opened 2023-01-11
Clear console

itzCozi:
Clearing the console is a task everyone has to do at-least once so why not put it here where lots of people will be able to use it.


ericklarac:
What would be the improvement of this?

itzCozi:
I found it very helpful when making CLI's as to not overload the user with to much text on one screen instead printing a part of that text say a paragraph then clearing the screen to give users a break.

PR #890, opened 2023-01-14
Added a new algorithm within strings module

Thejas-1:
So I came across this interesting problem that requires us to swap just one character amongst the two strings that are given to us. Once the swap is complete, the strings should contain the same number of distinct characters (The characters need not be the same in the two strings). The function should return True if it is possible to make such a swap and False otherwise.
For Example, if we are given two strings 'ddef' and 'dde', we can swap characters f from string 1 and d from string 2. Hence, our final strings will be 'dded' and 'fde', which will contain three distinct characters each (the duplicate characters count as 1) and the algorithm should return True. 
On the other hand, if we are given two strings 'cd' and 'e', irrespective of swapping characters c and d from string 1 with the character e in string 2, we will always have two distinct characters in string 1 and 1 character in string 2. Hence, the algorithm returns False.


keon:

PR #891, opened 2023-01-15
Python implementation for greedy graph colouring algo

abhaysinghraihal:


keon:
Thanks for the idea! The implementation needs more work (no tests, style issues). Closing, but graph coloring may be added in the future.

PR #904, opened 2023-09-01
Delete max_ones_index.py

darrenltj:
Deleted content in max_ones_index.py


keon:
Closing — this deletes an existing algorithm file without replacement.

PR #905, opened 2023-09-24
Add remove duplicates

oDqnger:
This algorithm simply removes any duplicates from an array.


tests/test_array.py
```diff
@@ -9,6 +9,7 @@
     missing_ranges,
     move_zeros,
     plus_one_v1, plus_one_v2, plus_one_v3,
+    remove_duplicates
```
evtn, line 12: you forgot a comma here

PR #909, opened 2023-11-25
fix: test josephus has add intiger list but passing string list

Krupeshgithub:
> **_algorithms/arrays/josephus_** has getting two parameters first one is int_list and second is skip. 
> **_tests/test_arrays/TestJosephus_**  there are passing list inside string values.

> - Just changing the ( **_tests/test_arrays/TestJosephus_** ) variable name with annotations and convert into list[int] or list[str].


keon:
Thanks! The test suite was overhauled in v1.0.0. Closing as stale.

PR #915, opened 2024-01-20
Added extended kd tree implemenatation

gjones1077:
Saw avl, bst and others so I decided to add my extended kd tree implementation. I put the testing file in the same place for bookkeeping purposes.


keon:

gjones1077:
Thank you for the update

PR #918, opened 2024-04-14
Fix remove_duplicates function to handle binary values

ca20110820:
This PR addresses an issue (originally found in unit test run) where the binary values for `1` and `True` (`0` and `False`) are treated equally for this function.

## Changes
- Updated implementation of `remove_duplicates(array)` function.
- Updated the unit tests for `remove_duplicates(array)` function.
- Added docstring to reflect the improved behavior and provide clear examples.


keon:
Thanks for the bug fix! The repo was restructured in v1.0.0. Closing as the changes no longer apply cleanly.

PR #919, opened 2024-04-14
Fixed minor bugs from unix.path

ca20110820:
Resolves minor bug seen from `tests/test_unix.py` under `TestUnixPath`.

The original implementation uses `"{}/{}"` which causes the bug when tested under Windows OS. I generalized it with:
```python
os.path.join('{}', '{}')
```
without assuming the local machine's OS.

## Changes
- Fixed error from `TestUnixPath.test_full_path` method with `os.path.join('{}', '{}')`.
- Fixed `unix.path.simplify_path.simplify_path_v1()` function with `posixpath.normpath()` function.

### DIFF
--- a/algorithms/unix/path/simplify_path.py
+++ b/algorithms/unix/path/simplify_path.py
@@ -14,10 +14,12 @@ In this case, you should ignore redundant slashes and return "/home/foo".
 
 Reference: https://leetcode.com/problems/simplify-path/description/
 """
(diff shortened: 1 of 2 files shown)


keon:
Thanks for the cross-platform fix! The repo was restructured in v1.0.0 with full linting. Closing as the changes no longer apply cleanly.

PR #930, opened 2024-10-18
Add compute_z_array Function with Unit Tests

Simranstha045:
### PR Description:

This PR introduces the implementation of the `compute_z_array` function, which computes the Z-array for a given string in linear time, as well as comprehensive unit tests to verify its correctness.

### Description of `compute_z_array` Function:
The `compute_z_array` function calculates the Z-array for a string, where each element at index `i` represents the length of the longest substring starting from index `i` that is also a prefix of the string. The algorithm runs in **O(n)** time by efficiently leveraging already computed values to avoid redundant comparisons.

**Key Features:**
- Utilizes the Z-box technique to reduce unnecessary comparisons.
- Processes the string in linear time.

### Unit Tests for `compute_z_array`:
The following tests have been added to ensure robustness:
- **Basic Test Cases:** Verifies correct Z-array generation for strings with repeating patterns (e.g., "aaaaa" and "ababa").
- **Edge Case Tests:**
  - Strings with no prefix matches (e.g., "abcde").
  - Full match strings (e.g., "aaaa").
  - Mixed case strings with partial matches.
  - Single character strings.
  - Empty string input.

These unit tests ensure the function works as expected across a range of inputs, providing confidence in its correctness and edge-case handling.

All tests have been added under the `unittest` framework to maintain consistency with the existing test suite.

### DIFF
--- a/algorithms/strings/__init__.py
+++ b/algorithms/strings/__init__.py
@@ -39,3 +39,4 @@ from .atbash_cipher import *
 from .longest_palindromic_substring import *
 from .knuth_morris_pratt import *
 from .panagram import *
+from .z_algorithm import *
--- a/algorithms/strings/z_algorithm.py
+++ b/algorithms/strings/z_algorithm.py
@@ -1,0 +1,36 @@
+"""
+Algorithm that computes the Z-array for a given string using the Z Algorithm.
+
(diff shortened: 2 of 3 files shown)


keon:

PR #2745, opened 2026-03-19
Optimize Quick Sort, N-Sum, and fix bug in Is Palindrome

DhruvrajSinhZala24:
1. Quick Sort: Added Median-of-Three pivot selection and Insertion Sort for small subarrays.\n2. N-Sum: Refactored to reduce redundant sorts and copies using recursion with indices.\n3. Is Palindrome: Fixed potential IndexError and improved efficiency.

### DIFF
--- a/algorithms/array/n_sum.py
+++ b/algorithms/array/n_sum.py
@@ -57,71 +57,46 @@ def n_sum(
     def _same_closure_default(a: Any, b: Any) -> bool:
         return a == b
 
-    def _n_sum_inner(n: int, nums: list[Any], target: Any) -> list[list[Any]]:
-        if n == 2:
-            results = _two_sum(nums, target)
-        else:
-            results = []
-            prev_num = None
-            for index, num in enumerate(nums):
-                if prev_num is not None and same_closure(prev_num, num):
-                    continue
-
-                prev_num = num
-                n_minus1_results = _n_sum_inner(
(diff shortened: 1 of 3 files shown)


keon (commented):
Verified locally: `test_array.py`, `test_sorting.py`, and `test_string.py` all pass (116/116). Algorithmically the changes look fine — median-of-three + insertion-sort threshold for quicksort is a standard, well-motivated optimization, and the n-sum refactor to indices rather than slicing is a real improvement.

A few process notes for the maintainer:

1. **Bundling.** Three unrelated changes (quicksort optimization, n-sum refactor, is_palindrome bugfix) in a single PR makes review and revert harder. Splitting into three PRs would be much friendlier to review.

2. **Overlap with #2749.** The is_palindrome portion duplicates #2749, which is a tighter, more focused fix to the same root cause (`_remove_punctuation` stripping digits). I'd suggest landing #2749 first and dropping the is_palindrome changes from this PR.

3. **n-sum behavior.** The refactor drops the `sorted([...])` call on each appended tuple that the original `_append_elem_to_each_list` had. The new code relies on the input being sorted (which it is, via the upstream `nums.sort()`) so output order is preserved — that's correct, but worth a one-line comment in the code to flag the invariant for future readers.

Not blocking — just suggesting a cleaner split.

PR #2749, opened 2026-05-06
fix: _remove_punctuation strips digits causing false palindrome results

faizzyhon:
## Problem

`_remove_punctuation()` in `algorithms/string/is_palindrome.py` used `ascii_letters` (a-z, A-Z only) to filter characters, silently discarding all digit characters (0-9).

Four of the five palindrome-check variants — `is_palindrome_reverse`, `is_palindrome_two_pointer`, `is_palindrome_stack`, and `is_palindrome_deque` — all call this helper, so they all returned wrong results for strings containing digits:

```python
# Before this fix:
is_palindrome_reverse("a1b2a")    # True  ← WRONG  (digits stripped → "aba" = palindrome)
is_palindrome_reverse("1a2")      # True  ← WRONG  (digits stripped → "a"   = palindrome)
is_palindrome("a1b2a")            # False ← correct (uses .isalnum() directly)
```

## Fix

Replace the `ascii_letters` membership test in `_remove_punctuation` with `str.isalnum()`, which correctly retains digit characters:

```python
- return "".join(char.lower() for char in text if char in ascii_letters)
+ return "".join(char.lower() for char in text if char.isalnum())
```

Remove the now-unused `from string import ascii_letters` import. Update all five functions' docstrings to say "alphanumeric" (was "alphabetic"). Add bound-guards `left < right` to the two inner while-loops in `is_palindrome` to prevent an infinite loop on all-punctuation input.

## Tests

Added `test_is_palindrome_with_digits` which exercises all five variants with:
- `"12321"` → `True` (valid digit palindrome)
- `"a1b2a"` → `False` (not a palindrome once digits are kept)

All 567 existing tests continue to pass.

### DIFF
--- a/algorithms/string/is_palindrome.py
+++ b/algorithms/string/is_palindrome.py
@@ -14,11 +14,12 @@ Complexity:
 from __future__ import annotations
 
 from collections import deque
-from string import ascii_letters
 
 
 def is_palindrome(text: str) -> bool:
     """Check if a string is a palindrome using two pointers on the original.
+
+    Considers only alphanumeric characters and ignores case.
 
     Args:
         text: The input string to check.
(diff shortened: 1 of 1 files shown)


keon (approved):
APPROVED. Verified locally: `test_string.py` 68/68 pass. Good catch — replacing the `ascii_letters` membership test with `str.isalnum()` matches the behavior of the standalone `is_palindrome` function (which already used `.isalnum()`), so all five variants now agree. The `left < right` bound-guards on the inner while-loops in `is_palindrome` are a nice incidental hardening against all-punctuation input.

Note for the maintainer: this overlaps with the is_palindrome portion of #2745.

PR #2750, opened 2026-05-06
fix: roman_to_int raises IndexError on empty string

faizzyhon:
## Problem

`roman_to_int('')` raises `IndexError: string index out of range` because the function unconditionally accesses `text[-1]` after the main loop, with no guard for an empty input string.

```python
# Before this fix:
roman_to_int('')  # IndexError: string index out of range
```

## Fix

Added an early-return guard at the top of the function:

```python
if not text:
    return 0
```

Also extended the docstring with `IV`, `MCMXCIV`, and `""` examples to document edge-case behavior.

## Tests

Added to `tests/test_string.py`:
- `test_roman_to_int_empty_string` — verifies `roman_to_int('')` returns `0`
- `test_roman_to_int_standard_cases` — verifies `III`, `IV`, `IX`, `XL`, `DCXXI`, `MCMXCIV`

### DIFF
--- a/algorithms/string/roman_to_int.py
+++ b/algorithms/string/roman_to_int.py
@@ -7,8 +7,8 @@ to be within the range from 1 to 3999.
 Reference: https://en.wikipedia.org/wiki/Roman_numerals
 
 Complexity:
-    Time:  O(n) where n is the length of the Roman numeral string
-    Space: O(1)
+Time: O(n) where n is the length of the Roman numeral string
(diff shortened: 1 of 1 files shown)


keon (approved):
APPROVED. Verified locally: `test_string.py` roman tests 2/2 pass. Early-return guard is the right shape and the added `IV` / `MCMXCIV` / `""` docstring examples are a nice touch.

Issue #2748, opened 2026-05-01
Add Chebyshev distance


rrbharath:
Add [Chebyshev distance](https://en.wikipedia.org/wiki/Chebyshev_distance) which is the maximum absolute difference between the coordinates

`Chebyshev Distance = max(|x1 - y1|, |x2 - y2|, ..., |xn - yn|)`

example:

x = (1, 2, 3)
y = (4, 6, 3)
Chebyshev Distance = max(|1–4|, |2–6|, |3–3|) = max(3, 4, 0) = 4

keon:
PR #2751 implements this. Currently open — pending review.

keon:
Implemented in #2751 (merged).

PR #2751, opened 2026-05-08
feat(math): add chebyshev distance

rrbharath:
Add chebyshev distance. Addresses https://github.com/keon/algorithms/issues/2748

### DIFF
--- a/algorithms/math/__init__.py
+++ b/algorithms/math/__init__.py
@@ -12,6 +12,7 @@ from algorithms.math import (
     modular_inverse,  # type: ignore[no-redef]  # noqa: E402
 )
 from algorithms.math.base_conversion import base_to_int, int_to_base
+from algorithms.math.chebyshev_distance import chebyshev_distance
 from algorithms.math.combination import combination, combination_memo
 from algorithms.math.cosine_similarity import cosine_similarity
 from algorithms.math.decimal_to_binary_ip import (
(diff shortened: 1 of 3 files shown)


keon (changes requested):
Verified locally: this PR's own tests fail.

```
tests/test_math.py::TestChebyshevDistance::test_chebyshev_distance
    TypeError: 'module' object is not callable
```

`test_math.py` imports `chebyshev_distance` from `algorithms.math`, but `algorithms/math/__init__.py` was not updated to re-export the function — so the import resolves to the *module* `algorithms.math.chebyshev_distance` rather than the function inside it. Calling it then fails.

Fix: add an explicit import (and matching `__all__` entry if one is maintained) in `algorithms/math/__init__.py`, e.g.

```python
from algorithms.math.chebyshev_distance import chebyshev_distance
```

For reference, see how the sibling `distance_between_two_points` and `manhattan_distance` are wired up in that same `__init__.py` — please follow the same pattern.

The algorithm itself looks correct.

keon:
I've pushed a commit to this branch that:

1. Adds the missing `chebyshev_distance` import and `__all__` entry to `algorithms/math/__init__.py` — this is what was breaking `test_chebyshev_distance` (`'module' is not callable`).
2. Polishes `algorithms/math/chebyshev_distance.py` to match the style of the sibling `manhattan_distance.py`: module docstring, `from __future__ import annotations`, type hints, doctest examples, and the project's `msg = ...; raise ValueError(msg)` convention.

Verified locally: `pytest tests/test_math.py -k Chebyshev` → 1 passed.

Issue #2753, opened 2026-05-13
Add geometric mean


rrbharath:

keon:
PR #2754 implements this. Currently open — pending review.

keon:
Implemented in #2754 (merged).

PR #2754, opened 2026-05-13
feat(math): add geometric mean

rrbharath:
Add geometric mean

Addresses issue: https://github.com/keon/algorithms/issues/2753

### DIFF
--- a/algorithms/math/__init__.py
+++ b/algorithms/math/__init__.py
@@ -38,6 +38,7 @@ from algorithms.math.generate_strobogrammtic import (
     gen_strobogrammatic,
     strobogrammatic_in_range,
 )
+from algorithms.math.geometric_mean import geometric_mean, geometric_mean_logarithm
 from algorithms.math.goldbach import goldbach, verify_goldbach
 from algorithms.math.is_strobogrammatic import is_strobogrammatic, is_strobogrammatic2
 from algorithms.math.krishnamurthy_number import krishnamurthy_number
@@ -93,6 +94,8 @@ __all__ = [
     "trailing_zero",
     "gen_strobogrammatic",
     "strobogrammatic_in_range",
+    "geometric_mean",
+    "geometric_mean_logarithm",
     "hailstone",
     "is_strobogrammatic",
     "is_strobogrammatic2",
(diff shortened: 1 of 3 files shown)


keon (changes requested):
Verified locally: this PR's own tests fail.

```
tests/test_math.py::TestGeometricMean::test_geometric_mean_logarithm_basic
    AssertionError: 100.0 != 100.00000000000004
```

`geometric_mean_logarithm([100, 100, 100])` returns `100.00000000000004` due to float drift through `math.log`/`math.exp`, so `assertEqual` rejects it. Same issue applies to `test_geometric_mean_basic` for any non-trivial input.

Fix: use `assertAlmostEqual(..., places=N)` for the exact-integer assertions, matching the style already used in `test_geometric_mean_list`. For example:

```python
def test_geometric_mean_logarithm_basic(self):
    self.assertAlmostEqual(4.0, geometric_mean_logarithm([2, 8]), places=10)
    self.assertAlmostEqual(100.0, geometric_mean_logarithm([100, 100, 100]), places=10)
```

Minor follow-ups (non-blocking):
- Consider exporting both functions in the public `algorithms.math` namespace via `__init__.py` so users can `from algorithms.math import geometric_mean` (the test does this and works because of star export, but it's worth being explicit).
- The two functions duplicate the empty/positive validation — extracting a small `_validate(numbers)` helper would DRY it up.

keon:
I've pushed a commit to this branch that:

1. Replaces `assertEqual` with `assertAlmostEqual(places=10)` in `test_geometric_mean_basic` and `test_geometric_mean_logarithm_basic` — the previous assertions failed because the cube root of 1e6 is not exactly representable in IEEE-754 (`geometric_mean([100,100,100]) == 99.99999999999997`, `geometric_mean_logarithm([100,100,100]) == 100.00000000000004`).
2. Fixes the misleading `>>> geometric_mean([100, 100, 100]) -> 100.0` doctest line in `geometric_mean.py` by wrapping it in `round(..., 4)` like the sibling examples — it would have failed under `pytest --doctest-modules` for the same reason.

Verified locally: `pytest tests/test_math.py -k Geometric` → 10 passed.

PR #2759, opened 2026-05-24
Fix: tighten loop bound in first_switch_pairs

vdl299:
## Summary

* Tighten loop bound in `first_switch_pairs`
* Remove redundant empty-check guard made unnecessary by the fix

## Improvement

`range(len(storage_stack))` was a misleading loop bound — the loop actually
ran at most `len/2` iterations, relying on internal `break` guards to terminate
correctly. Replaced with `(len(storage_stack)+1)//2` which reflects the exact
number of iterations needed. The first empty-check guard became redundant and
was removed.

## Test plan

* `python -m pytest tests/test_stack.py` - 10 tests pass


keon (approved):
thanks for the contribution!

